PR #190, opened 2020-01-08
[Type-o-Matic] StoreKit

xxaxdxcxx:
Adds type name maps for StoreKit


mandel-macaque:
LGTM and Makefile is a lot nicer now in master.

PR #192, opened 2020-01-08
[Type-o-Matic] HomeKit and alphabetization

xxaxdxcxx:
Adds type name maps for namespace.

Note that in this case 6 enum types defined in the Xamarin iOS SDK are skipped; this is because there are no corresponding enums provided by Apple. Rather, they have pages in their documentation collecting all the global variables, but no enumeration. The pertinent enums and their corresponding Apple documentation pages are [HMAccessoryCategoryType](https://developer.apple.com/documentation/homekit/hmaccessorycategory/accessory_category_types), [HMActionSetType](https://developer.apple.com/documentation/homekit/hmactionset/action_set_types), [HMCharacteristicMetadataFormat](https://developer.apple.com/documentation/homekit/hmcharacteristicmetadata/characteristic_data_formats), [HMCharacteristicMetadataUnits](https://developer.apple.com/documentation/homekit/hmcharacteristicmetadata/characteristic_units), [HMCharacteristicType](https://developer.apple.com/documentation/homekit/hmcharacteristic/characteristic_types), and [HMServiceType](https://developer.apple.com/documentation/homekit/hmservice/accessory_service_types).

### DIFF
--- a/SwiftReflector/Importing/TypeAggregator.iOS.cs
+++ b/SwiftReflector/Importing/TypeAggregator.iOS.cs
@@ -40,48 +40,20 @@ namespace SwiftReflector.Importing {
 
 		static partial void TypesToSkipIOS (ref HashSet<string> result) { result = iOSTypesToSkip; }
 		static HashSet<string> iOSTypesToSkip = new HashSet<string> () {
-			"Messages.MSMessagesAppPresentationContext",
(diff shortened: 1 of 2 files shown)


mandel-macaque:
@xxaxdxcxx that is probably due to the fact that we use smart enums: https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/binding-types-reference

PR #194, opened 2020-01-08
[Type-o-Matic] PdfKit

xxaxdxcxx:
Adds type name maps for PdfKit.

Note that the differences in capitalization are essential; in the `iOSTypesToSkip` set it must be written `PdfKit`, but in the `iOSTypeNamesToMap` dictionary it must be written `PDFKit`.


mandel-macaque:
@xxaxdxcxx correct, that is because in C# we use the C# style convention for naming which does not match the one used in objc.

PR #222, opened 2020-01-13
Added some Associated Type tools

stephen-hawley:
Added a predicate and a tool to find an associated type by name.

Also fixed a warning on `GatherXObjects` and implemented `CompleteUnrooting`.

As a side note, when we emit XML type binding information, we go through a process of unrooting which eliminates all (potential) circular references in the type. Primarily, this is like a copy constructor that rips out the `Parent` pointer but leaves enough information to reconstruct it later.

Since it's a copier, we should actually, you know, copy things.

### DIFF
--- a/SwiftReflector/SwiftXmlReflection/ProtocolDeclaration.cs
+++ b/SwiftReflector/SwiftXmlReflection/ProtocolDeclaration.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System.Collections.Generic;
+using System.Linq;
 using System.Xml.Linq;
 
 namespace SwiftReflector.SwiftXmlReflection {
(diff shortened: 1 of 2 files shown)


SwiftReflector/SwiftXmlReflection/ProtocolDeclaration.cs
```diff
@@ -31,7 +32,22 @@ protected virtual void GatherXObjects (List<XObject> xobjects)
 			xobjects.Add (new XElement ("associatedtypes", assocTypes.ToArray ()));
 		}
 
+		protected override void CompleteUnrooting (TypeDeclaration unrooted)
+		{
+			base.CompleteUnrooting (unrooted);
+			if (unrooted is ProtocolDeclaration pd) {
```
mandel-macaque: 👍

PR #249, opened 2020-01-16
[Type-o-Matic] MediaToolbox

xxaxdxcxx:
Adds type name maps for MediaToolbox

### DIFF
--- a/SwiftReflector/Importing/TypeAggregator.iOS.cs
+++ b/SwiftReflector/Importing/TypeAggregator.iOS.cs
@@ -217,6 +217,8 @@ namespace SwiftReflector.Importing {
 			"MapKit.MKDirectionsMode", // not an enum
 			// MediaPlayer
 			"MediaPlayer.MPMovieControlMode", // not an enum
+			// MediaToolbox
+			"MediaToolbox.MTAudioProcessingTapError", // can't find it
 			// Messages
 			"Messages.MSMessagesAppPresentationContext",
 			// PassKit
--- a/swiftglue/Makefile
+++ b/swiftglue/Makefile
@@ -54,6 +54,7 @@ IOS_NAMESPACES= \
 	MapKit \
 	MediaAccessibility \
 	MediaPlayer \
+	MediaToolbox \
 	PassKit \
 	PdfKit \
 	PushKit \


SwiftReflector/Importing/TypeAggregator.iOS.cs
```diff
@@ -217,6 +217,8 @@ public partial class TypeAggregator {
 			"MapKit.MKDirectionsMode", // not an enum
 			// MediaPlayer
 			"MediaPlayer.MPMovieControlMode", // not an enum
+			// MediaToolbox
+			"MediaToolbox.MTAudioProcessingTapError", // can't find it
```
whitneyschmidt: Could you add more of an explanation re `can't find it`? I'm not super familiar with swift so this might be obvious :)

xxaxdxcxx: @whitneyschmidt So basically all I'm doing is finding enums in Xamarin's .NET API whose names don't match corresponding enums in Swift's API. The problem is that not every Xamarin enum seems to have a corresponding Swift enum. In some cases, this is because the Xamarin enum is actually just a collection of constants in Swift, documented on the same page and given similar names but not actually part of an enum or struct. In other cases, it seems like there is no corresponding enum/struct. I'm pretty sure this is due to Apple's API changing, but unfortunately they don't document those changes on their website. Also sometimes they make enums anonymous, which makes them impossible to find with search. So I put `// can't find it` if after searching and digging through related documentation for a while, I can't find any reference to such an enum/struct.

So in this example, the struct ([here's the Xamarin documentation](https://docs.microsoft.com/en-us/dotnet/api/mediatoolbox.mtaudioprocessingtaperror?view=xamarin-ios-sdk-12)) does not show up in [Apple's documentation](https://developer.apple.com/search/?q=MTAudioProcessingTapError). And furthermore neither does the class it relates to (see [this effectively empty search](https://developer.apple.com/search/?q=MTAudioProcessingTap)). So it's unclear (to me) what this could even bind it to, so I put it in the list to skip along with that comment.

For what it's worth, there were examples in the source code from before my time that made me realize this was possible.

Hope this answers your question!

whitneyschmidt: Thanks for the detailed response! That answered my question and I also understand a lot more about your project now 🎊

PR #287, opened 2020-01-21
[Type-o-Matic] CoreMedia

xxaxdxcxx:
Adds type name maps for CoreMedia.

### DIFF
--- a/SwiftReflector/Importing/TypeAggregator.iOS.cs
+++ b/SwiftReflector/Importing/TypeAggregator.iOS.cs
@@ -167,6 +167,16 @@ namespace SwiftReflector.Importing {
 			"CoreImage.CIImageOrientation", // not in this namespace
 			"CoreImage.CIWrapMode", // can't find it
 			"CoreImage.FaceDetectorAccuracy", // not an enum
+			// CoreMedia
+			"CoreMedia.CMFormatDescriptionError", // not an enum
+			"CoreMedia.CMPixelFormat", // not an enum
+			"CoreMedia.CMTimebaseError", // not an enum
+			"CoreMedia.LensStabilizationStatus", // can't find it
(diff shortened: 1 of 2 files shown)


SwiftReflector/Importing/TypeAggregator.iOS.cs
```diff
@@ -477,6 +487,11 @@ public partial class TypeAggregator {
 			// CoreImage
 			{ "CoreImage.CIDataMatrixCodeEccVersion", "CIDataMatrixCodeDescriptor.ECCVersion" },
 			{ "CoreImage.CIQRCodeErrorCorrectionLevel", "CIQRCodeDescriptor.ErrorCorrectionLevel" },
+			// CoreMedia
+			// iOS 13.0 or later: { "CoreMedia.CMBlockBufferError", "CMBlockBuffer.Error" },
+			// iOS 13.0 or later: { "CoreMedia.CMClockError", "CMClock.Error" },
+			// iOS 13.0 or later: { "CoreMedia.CMSampleBufferError", "CMSampleBuffer.Error" },
+			// iOS 13.0 or later: { "CoreMedia.CMSyncError", "CMSync.Error" },
```
mandel-macaque: @stephen-hawley is this the correct approach?

xxaxdxcxx: To add context (not sure if this is correct or not either): these were the replacements suggested by Swift's errors, but when I tried to compile with these replacements it threw the version requirement. So I left the replacements in as comments for the time being (to make it easier for if someone goes through and updates these for iOS 13) and put them in namespaces to ignore.

stephen-hawley: 👍 We're not compiling to 13.0 (yet)

PR #296, opened 2020-01-22
[Type-o-Matic] Fixed duplicate EKWeekday declaration

xxaxdxcxx:
Originally the `EKDay` enum had been mapped to the `EKWeekday` enum - however the `EKWeekday` enum already existed, so this was a duplicate binding. So this change just skips the `EKDay` enum, which is an artifact of Objective-C and not used in Swift.

### DIFF
--- a/SwiftReflector/Importing/TypeAggregator.iOS.cs
+++ b/SwiftReflector/Importing/TypeAggregator.iOS.cs
@@ -196,6 +196,7 @@ namespace SwiftReflector.Importing {
 			"CoreVideo.CVPixelFormatType", // not an enum
 			// EventKit
 			"EventKit.EKCalendarEventAvailability", // not an enum
+			"EventKit.EKDay", // changed to EKWeekday
 			// Foundation
 			"Foundation.NSFileType",
 	    		"Foundation.NSUserDefaultsType",
@@ -515,7 +516,6 @@ namespace SwiftReflector.Importing {
 			// CoreVideo
 			{ "CoreVideo.CVPixelBufferLock", "CVPixelBufferLockFlags" },
 			// EventKit
-			{ "EventKit.EKDay", "EKWeekday" },
 			{ "EventKit.EKErrorCode", "EKError.Code" },
 	    		// Foundation
 			{ "Foundation.NSBundle", "Bundle" },


type-o-matic/Program.cs
```diff
@@ -286,6 +286,11 @@ static void TLFunctionsForStream (Stream stm, PlatformName platform, Dictionary<
 
 		static SLAttribute AvailableAttributeForFunc (PlatformName platform, TypeDefinition type)
 		{
+
+			var map = TypeAggregator.AvailableMapForPlatform (platform);
+			string available;
+			if (map != null && map.TryGetValue (type.FullName, out available))
```
mandel-macaque: With the new c# syntax you can remove string available and do something like:
```suggestion
			if (map != null && map.TryGetValue (type.FullName, out string available))
```

SwiftReflector/MarshalEngineSwiftToCSharp.cs
```diff
@@ -32,8 +32,7 @@ public IEnumerable<ICodeElement> MarshalFunctionCall (FunctionDeclaration func,
 			postMarshalCode.Clear ();
 			ICodeElement returnLine = null;
 			var instanceEntity = typeMapper.GetEntityForTypeSpec (func.ParameterLists [0] [0].TypeSpec);
-			bool instanceIsProtocol = instanceEntity.EntityType == EntityType.Protocol;
-			//			bool returnIsGeneric = func.ReturnTypeSpec != null && func.IsGenericType(func.ReturnTypeSpec);
+			bool instanceIsProtocol = instanceEntity != null && instanceEntity.EntityType == EntityType.Protocol;
```
mandel-macaque: You do not need to check if is null, you can do something like:
```suggestion
			bool instanceIsProtocol = instanceEntity?.EntityType == EntityType.Protocol;
```

The above like, if instanceEntity is null, will return null for EntityType which is not == to Protocol.

SwiftReflector/OverrideBuilder.cs
```diff
@@ -191,6 +199,32 @@ ExtensionDeclaration BuildExtensionDefinition (ModuleDeclaration targetModule)
 			return decl;
 		}
 
+		ClassDeclaration BuildAssociatedTypeOverride (string name, ModuleDeclaration targetModule)
+		{
+			name = name ?? ProxyClassName (OriginalClass);
+			var protocol = OriginalClass as ProtocolDeclaration;
```
mandel-macaque: You are doing and as, yet not check for protocol to be null. We might have issues.

SwiftReflector/OverrideBuilder.cs
```diff
@@ -191,6 +199,32 @@ ExtensionDeclaration BuildExtensionDefinition (ModuleDeclaration targetModule)
...
+
+			var decl = new ClassDeclaration ();
+			decl.Name = name;
+			decl.Module = targetModule;
+			foreach (var at in protocol.AssociatedTypes) {
```
mandel-macaque: As mentioned above, protocol can be null.

SwiftReflector/OverrideBuilder.cs
```diff
@@ -191,6 +199,32 @@ ExtensionDeclaration BuildExtensionDefinition (ModuleDeclaration targetModule)
...
+					if (at.ConformingProtocols.Count == 1) {
+						genDecl.Constraints.Add (new InheritanceConstraint (genName, at.ConformingProtocols [0]));
+					} else {
+						genDecl.Constraints.Add (new InheritanceConstraint (genName, new ProtocolListTypeSpec (at.ConformingProtocols)));
+					}
+				}
+				decl.Generics.Add (genDecl);
+			}
+			decl.Inheritance.Add (new Inheritance (OriginalClass.ToFullyQualifiedName (), InheritanceKind.Class));
+			return decl.MakeUnrooted () as ClassDeclaration;
```
mandel-macaque: Is null a valid return result here? @stephen-hawley

SwiftReflector/OverrideBuilder.cs
```diff
@@ -221,10 +255,18 @@ void CopyVirtualMethods ()
 			else
 				HandleSuperClassVirtualMethods (OriginalClass);
 			IndexOfFirstNewVirtualMethod = OverriddenVirtualMethods.Count;
-			OverriddenVirtualMethods.AddRange (VirtualMethodsForClass (OriginalClass).Select (m => MarkOverrideSurrogate (m, Reparent (new FunctionDeclaration (m), OverriddenClass))));
-			var members = isProtocol ? EveryProtocolExtension.Members : OverriddenClass.Members;
+			if (isProtocol && hasAssociatedTypes) {
+				OverriddenVirtualMethods.AddRange (VirtualMethodsForClass (OriginalClass).Select (m => MarkOverrideSurrogate (m, Reparent (RebuildFunctionDeclarationWithAssociatedTypes (OverriddenClass, m), OverriddenClass))));
+			} else {
+				OverriddenVirtualMethods.AddRange (VirtualMethodsForClass (OriginalClass).Select (m => MarkOverrideSurrogate (m, Reparent (new FunctionDeclaration (m), OverriddenClass))));
+			}
```
mandel-macaque: This is personal, but for debugging, I would create a range var, add the result of the LINQ and then do a single call to AddRange.

```csharp
IEnumberable range;
if (isProtocol && hasAssociatedTypes) {
 				range = VirtualMethodsForClass (OriginalClass).Select (m => MarkOverrideSurrogate (m, Reparent (RebuildFunctionDeclarationWithAssociatedTypes (OverriddenClass, m), OverriddenClass))));
 			} else {
 				range = VirtualMethodsForClass (OriginalClass).Select (m => MarkOverrideSurrogate (m, Reparent (new FunctionDeclaration (m), OverriddenClass))));
 			}
OverriddenVirtualMethods.AddRange (range);
```csharp
When we debug, we will be able to see the range value, will be simpler.

File: SwiftReflector/OverrideBuilder.cs [hunk 7]
```diff
@@ -221,10 +255,18 @@ void CopyVirtualMethods ()
...
+			var members = isProtocol && !hasAssociatedTypes ? EveryProtocolExtension.Members : OverriddenClass.Members;
 			members.AddRange (OverriddenVirtualMethods);
-			members.AddRange (VirtualPropertiesForClass (OriginalClass).Select (p => Reparent (new PropertyDeclaration (p), OverriddenClass)));
+			if (isProtocol && hasAssociatedTypes) {
+				members.AddRange (VirtualPropertiesForClass (OriginalClass).Select (p => Reparent (RebuildPropertyDeclaration (p), OverriddenClass)));
+			} else {
+				members.AddRange (VirtualPropertiesForClass (OriginalClass).Select (p => Reparent (new PropertyDeclaration (p), OverriddenClass)));
+			}
```
mandel-macaque: Same as the above comment, we call AddRange with two diff values, lets add a var and then a single call.

SwiftReflector/OverrideBuilder.cs
```diff
@@ -314,17 +363,17 @@ void CreateSLImplementation ()
 							EqualityConstraint eq = bc as EqualityConstraint;
 							if (eq != null) {
 								SLType secondType = OriginalClass.IsTypeSpecGeneric (eq.Type2Spec) ?
-								                                 new SLSimpleType (GenericName (OriginalClass, eq.Type2)) :
-								                                 typeMapper.OverrideTypeSpecMapper.MapType (OriginalClass, Imports, eq.Type2Spec, false);
+												 new SLSimpleType (GenericName (OriginalClass, eq.Type2)) :
+												 typeMapper.OverrideTypeSpecMapper.MapType (OriginalClass, Imports, eq.Type2Spec, false);
```
mandel-macaque: Noise, please undo white space changes.

SwiftReflector/OverrideBuilder.cs
```diff
@@ -314,17 +363,17 @@ void CreateSLImplementation ()
...
 															   secondType);
 							} else {
 								var inh = bc as InheritanceConstraint;
 								if (inh == null)
 									throw ErrorHelper.CreateError (ReflectorError.kTypeMapBase + 40, $"Unexpected constraint type {bc.GetType ().Name}");
 								SLType secondType = OriginalClass.IsTypeSpecGeneric (inh.InheritsTypeSpec) ?
-								                                 new SLSimpleType (GenericName (OriginalClass, inh.Inherits)) :
-								                                 typeMapper.OverrideTypeSpecMapper.MapType (OriginalClass, Imports, inh.InheritsTypeSpec, false);
+												 new SLSimpleType (GenericName (OriginalClass, inh.Inherits)) :
+												 typeMapper.OverrideTypeSpecMapper.MapType (OriginalClass, Imports, inh.InheritsTypeSpec, false);
```
mandel-macaque: Same, lets try to remove as much noise as possible.

mandel-macaque:
Looks like I reviewed code from @stephen-hawley  and not from @xxaxdxcxx hehe

PR #299, opened 2020-01-24
[Type-o-Matic] Refactor Type-o-Matic calling process

xxaxdxcxx:
This PR is the first step towards making a more efficient Type-o-Matic implementation. It lays the groundwork by allowing Type-o-Matic to be called with namespaces, even when generating C#. To do this, it makes several changes:
1. Add another flag to the Type-o-Matic call, `--generate=(csharp|swift)`. This allows the user to expressly decide to create Swift or C# code, instead of relying on the other arguments to decide.
2. Pull the Type-o-Matic call from the SwiftRuntimeLibrary out of msbuild and into make. **This does not change the entire SwiftRuntimeLibrary build process to make, it just pulls one part into a makefile.** This is done to enable the next change:
3. Pull all included namespaces into `common.mk`, then include namespaces in the Type-o-Matic call from SwiftRuntimeLibrary. This means that we now have access to the pertinent namespaces when generating C# code with Type-o-Matic, which is necessary for the planned refactor.

The next step after this PR is to move to a hash table holding type conversions, instead of a single function per type conversion. This approach requires knowing about namespaces from the C# side, hence these changes.

Note: some of the changes in this PR come from running Edit -> Format -> Format Document in `Program.cs`.


type-o-matic/Program.cs
```diff
@@ -95,7 +104,7 @@ static CSFile GeneratePInvokesFromTypes (TypeAggregator types, PlatformName plat
 				.And (new CSUsing ("System"))
 				.And (new CSUsing ("System.Collections.Generic"))
 				.And (new CSUsing ("SwiftRuntimeLibrary"));
-		
+
```
mandel-macaque: Less try and do not add noise in the diff.

type-o-matic/Program.cs
```diff
@@ -129,7 +138,7 @@ static CSFile GeneratePInvokesFromTypes (TypeAggregator types, PlatformName plat
 					}
 				}
 			}
-			
+
```
mandel-macaque: Same, please no noise.

type-o-matic/Program.cs
```diff
@@ -147,8 +156,8 @@ static CSMethod PInvokeForType (TypeDefinition type, Dictionary<string, string>
 			if (!funcs.TryGetValue (funcID, out mangledName))
 				return null;
 
-	    		return CSMethod.PInvoke (CSVisibility.Internal, new CSSimpleType ("SwiftMetatype"), funcID, new CSIdentifier ("SwiftCore.kXamGlue"),
-					mangledName.Substring (1), new CSParameterList ());
+			return CSMethod.PInvoke (CSVisibility.Internal, new CSSimpleType ("SwiftMetatype"), funcID, new CSIdentifier ("SwiftCore.kXamGlue"),
+				    mangledName.Substring (1), new CSParameterList ());
```
mandel-macaque: Same.

type-o-matic/Program.cs
```diff
@@ -174,7 +183,7 @@ static SLFunc MetadataFuncForType (PlatformName platform, TypeDefinition type, T
 		static string FuncIDForTypeDefinition (TypeDefinition type)
 		{
 			// Used to be MetaDataWrapperFor
-	    		// The size of this symbol gets magnified by the number of types.
+			// The size of this symbol gets magnified by the number of types.
 			// Removing 14 characters saves ~32K
```
mandel-macaque: Same.

type-o-matic/Program.cs
```diff
@@ -206,7 +215,7 @@ static SLFile GenerateStubsFromTypes (TypeAggregator types, PlatformName platfor
 			return slfile;
 		}
 
-		static IEnumerable <SLFunc> MetaWrapperForFunc (PlatformName platform, IEnumerable<TypeDefinition> types,
+		static IEnumerable<SLFunc> MetaWrapperForFunc (PlatformName platform, IEnumerable<TypeDefinition> types,
```
mandel-macaque: Same.

type-o-matic/Program.cs
```diff
@@ -425,7 +434,7 @@ static string AttributeNameForPlatform (PlatformName platform)
 			}
 		}
 
-		static SLIdentifier PlatformCondition(PlatformName platform)
+		static SLIdentifier PlatformCondition (PlatformName platform)
```
mandel-macaque: Same.

type-o-matic/Program.cs
```diff
@@ -39,7 +39,7 @@ public static int Main (string [] args)
 				return 1;
 			}
 
-	    		if (options.Platform == PlatformName.None) {
+			if (options.Platform == PlatformName.None) {
```
mandel-macaque: Remove noise please.

SwiftRuntimeLibrary.Mac/SwiftRuntimeLibrary.Mac.csproj
```diff
@@ -284,8 +284,7 @@
     <Reference Include="System" />
   </ItemGroup>
   <Target Name="GeneratedCSCode" BeforeTargets="CoreCompile" Inputs="$(MSBuildProjectFullPath)" Outputs="GeneratedCode\BindingMetadata.MacOS.cs">
-    <Exec Command="mkdir -p GeneratedCode" />
-    <Exec Command="/Library/Frameworks/Mono.framework/Versions/Current/Commands/mono $(MSBuildProjectDirectory)\..\type-o-matic\bin\Debug\type-o-matic.exe --swift-lib-path=$(MSBuildProjectDirectory)/../apple/build/Ninja-ReleaseAssert/swift-macosx-x86_64/lib/swift/ --platform=mac --xamglue-framework=$(MSBuildProjectDirectory)/../swiftglue/bin/$(Configuration)/mac/FinalProduct/XamGlue.framework --output=GeneratedCode\BindingMetadata.MacOS.cs" />
+    <Exec Command="cd ../SwiftRuntimeLibrary ; make create-pinvokes-mac configuration=$(Configuration)" />
```
mandel-macaque: There is not need to do a `cd` here. Take a look at the -C option in make.

> -C dir, --directory=dir
            Change to directory dir before reading the makefiles or doing anything else.  If multiple -C options are specified, each is  interpreted  relative  to
            the previous one: -C / -C etc is equivalent to -C /etc.  This is typically used with recursive invocations of make.

SwiftRuntimeLibrary.iOS/SwiftRuntimeLibrary.iOS.csproj
```diff
@@ -281,8 +281,7 @@
     </Compile>
   </ItemGroup>
   <Target Name="GeneratedCSCode" BeforeTargets="CoreCompile" Inputs="$(MSBuildProjectFullPath)" Outputs="GeneratedCode\BindingMetadata.iOS.cs">
-    <Exec Command="mkdir -p GeneratedCode" />
-    <Exec Command="/Library/Frameworks/Mono.framework/Versions/Current/Commands/mono $(MSBuildProjectDirectory)\..\type-o-matic\bin\Debug\type-o-matic.exe --swift-lib-path=$(MSBuildProjectDirectory)/../apple/build/Ninja-ReleaseAssert/swift-macosx-x86_64/lib/swift/ --platform=iphone --xamglue-framework=$(MSBuildProjectDirectory)/../swiftglue/bin/$(Configuration)/iphone/FinalProduct/XamGlue.framework --output=GeneratedCode\BindingMetadata.iOS.cs" />
+    <Exec Command="cd ../SwiftRuntimeLibrary ; make create-pinvokes-ios configuration=$(Configuration)" />
```
mandel-macaque: Same, no need to do a cd.

SwiftRuntimeLibrary/Makefile
```diff
@@ -0,0 +1,24 @@
...
+	rm -rf $(TOP)/SwiftRuntimeLibrary.iOS/GeneratedCode/*
+
+clean-mac:
+	rm -rf $(TOP)/SwiftRuntimeLibrary.Mac/bin/*
+	rm -rf $(TOP)/SwiftRuntimeLibrary.Mac/GeneratedCode/*
+
+clean: clean-ios clean-mac
+	
+create-pinvokes-ios:
+	$(shell mkdir -p $(TOP)/SwiftRuntimeLibrary.iOS/GeneratedCode)
```
mandel-macaque: Why the use of shell??

SwiftRuntimeLibrary/Makefile
```diff
@@ -0,0 +1,24 @@
...
+	$(shell /Library/Frameworks/Mono.framework/Versions/Current/Commands/mono $(TOP)/type-o-matic/bin/Debug/type-o-matic.exe --swift-lib-path=$(TOP)/apple/build/Ninja-ReleaseAssert/swift-macosx-x86_64/lib/swift/ --platform=iphone --generate=csharp $(addprefix --namespace=, $(IOS_NAMESPACES)) --xamglue-framework=$(TOP)/swiftglue/bin/$(configuration)/iphone/FinalProduct/XamGlue.framework --output=$(TOP)/SwiftRuntimeLibrary.iOS/GeneratedCode/BindingMetadata.iOS.cs)
+
+create-pinvokes-mac:
+	$(shell mkdir -p $(TOP)/SwiftRuntimeLibrary.Mac/GeneratedCode)
+	$(shell /Library/Frameworks/Mono.framework/Versions/Current/Commands/mono $(TOP)/type-o-matic/bin/Debug/type-o-matic.exe --swift-lib-path=$(TOP)/apple/build/Ninja-ReleaseAssert/swift-macosx-x86_64/lib/swift/ --platform=mac --generate=csharp $(addprefix --namespace=, $(MAC_NAMESPACES)) --xamglue-framework=$(TOP)/swiftglue/bin/$(configuration)/mac/FinalProduct/XamGlue.framework --output=$(TOP)/SwiftRuntimeLibrary.Mac/GeneratedCode/BindingMetadata.MacOS.cs)
```
mandel-macaque: Same.

SwiftRuntimeLibrary/Makefile
xxaxdxcxx: @mandel-macaque copying the pattern [here](https://github.com/xamarin/binding-tools-for-swift/blob/master/swiftglue/Makefile#L148). What is the alternative? (sorry I'm not super familiar with makefiles)

xxaxdxcxx: When I try changing it to just `mkdir -p $(TOP)/SwiftRuntimeLibrary.iOS/GeneratedCode` (removing the `$(shell ...)`), the next command (calling Type-o-Matic) fails and says
```
Could not find a part of the path "/binding-tools-for-swift/SwiftRuntimeLibrary.iOS/GeneratedCode/BindingMetadata.iOS.cs".
```
But I can see the directory is being made; why would this happen?

common.mk
```diff
@@ -26,3 +26,136 @@ Q_LIPO       = $(if $(V),,@echo "LIPO   $(@F)";)
 Q_SOM        = $(if $(V),,@echo "BINDINGTOOLSFORSWIFT $(NAME)";)
 Q_XCODEBUILD = $(if $(V),,@echo "XCODEBUILD/$* $(NAME)";)
 Q_SWIFTC     = $(if $(V),,@echo "SWIFTC $(@F)";)
+
+# namespaces
+COMMON_NAMESPACES= \
```
stephen-hawley: Minor - since you're putting this into a global include, it's best to now make these identifiers self-identify more. Suggest `TYPE_O_MATIC_COMMON_NAMESPACES` and `TYPE_O_MATIC_IOS_NAMESPACES` and ``TYPE_O_MATIC_MAC_NAMESPACES`

type-o-matic/Program.cs
xxaxdxcxx: Done

SwiftRuntimeLibrary/Makefile
xxaxdxcxx: Resolved by removing `$(shell ...)` entirely; output is much nicer now!

common.mk
xxaxdxcxx: Done

type-o-matic/Program.cs
```diff
@@ -66,10 +75,10 @@ public static int Main (string [] args)
 			aggregatedTypes.Aggregate ();
 
 			var writer = new CodeWriter (options.OutputWriter);
-			if (options.Namespaces.Count > 0) {
+			if (options.ToGenerate.Equals ("swift")) {
```
mandel-macaque: Better to use this version of the method: https://docs.microsoft.com/en-us/dotnet/api/system.string.equals?view=netframework-4.8#System_String_Equals_System_String_System_String_System_StringComparison_

type-o-matic/Program.cs
```diff
@@ -51,12 +51,21 @@ public static int Main (string [] args)
...
 				return 1;
 			}
-			
+
+			if (options.Namespaces.Count == 0) {
+				Console.WriteLine ("You need to select input namespaces.");
+				return 1;
+			}
+
+			if (options.ToGenerate.Equals ("csharp") && options.Framework == null) {
```
mandel-macaque: Better to use this version of the method: https://docs.microsoft.com/en-us/dotnet/api/system.string.equals?view=netframework-4.8#System_String_Equals_System_String_System_String_System_StringComparison_

type-o-matic/Program.cs
xxaxdxcxx: Fixed!

type-o-matic/Program.cs
xxaxdxcxx: Fixed!

SwiftRuntimeLibrary/Makefile
```diff
@@ -0,0 +1,34 @@
...
+	@rm -rf $(TOP)/SwiftRuntimeLibrary.iOS/GeneratedCode
+
+clean-mac:
+	@echo "Cleaning binary files and generated code for Mac"
+	@rm -rf $(TOP)/SwiftRuntimeLibrary.Mac/bin/*
+	@rm -rf $(TOP)/SwiftRuntimeLibrary.Mac/GeneratedCode
+
+clean: clean-ios clean-mac
+	
+create-pinvokes-ios:
```
rolfbjarne: This is not the optimal way of writing a makefile, because the makefile will re-execute everything if you run `make` again, even if nothing changed.

The better way to do it is to identify the inputs and output of the task you want to accomplish. In this case it seems the output is `$(TOP)/$(BINDING_METADATA_IOS)`, but I'm not sure what the inputs are.

In any case, the format of the make rule would be something like this:

```make
$(TOP)/$(BINDING_METADATA_IOS): inputfile1 inputfile2 inputfileX
    @echo "Generating code for iOS..."
    @mkdir -p $(TOP)/SwiftRuntimeLibrary.iOS/GeneratedCode
    @/Library/Frameworks/Mono.framework/Versions/Current/Commands/mono $(TOP)/type-o-matic/bin/Debug/type-o-matic.exe --swift-lib-path=$(TOP)/apple/build/Ninja-ReleaseAssert/swift-macosx-x86_64/lib/swift/ --platform=iphone --generate=csharp $(addprefix --namespace=, $(TYPE_O_MATIC_IOS_NAMESPACES)) --xamglue-framework=$(TOP)/swiftglue/bin/$(configuration)/iphone/FinalProduct/XamGlue.framework --output=$(TOP)/$(BINDING_METADATA_IOS)
    @echo "Generated C# binding metadata files for iOS at $(TOP)/$(BINDING_METADATA_IOS)"

$(TOP)/$(BINDING_METADATA_MAC): ...
    ...

create-pinvokes: $(TOP)/$(BINDING_METADATA_IOS) $(TOP)/$(BINDING_METADATA_MAC)
```

mandel-macaque: @xxaxdxcxx please address this before we merge.

xxaxdxcxx: @rolfbjarne @mandel-macaque addressed in latest commit :)

PR #300, opened 2020-01-27
Added code for doing declarations of PATs

stephen-hawley:
This next step in the Protocol with Associated Types evolution: turns the PAT into a generic interface declaration.

Keeping the PR's smaller, there is no actual code generated for the wrappers and no members of the interface yet.

I also added in a constructor into the phony `ClassDeclaration` to ensure that that is available by wrapping code.

There is support for constraints in the class/interface declaration, but that's not tested yet.
The existing smoke test exercises all of this code. No smoke.

### DIFF
--- a/SwiftReflector/NewClassCompiler.cs
+++ b/SwiftReflector/NewClassCompiler.cs
@@ -1563,8 +1563,6 @@ namespace SwiftReflector {
 						errors.Add (ex);
 						continue;
 					}
-					if (decl.HasAssociatedTypes)
-						continue;
 					var use = new CSUsingPackages ("System", "System.Runtime.InteropServices");
 					string nameSpace = TypeMapper.MapModuleToNamespace (decl.Module.Name);
 					var nm = new CSNamespace (nameSpace);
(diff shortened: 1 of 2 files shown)


SwiftReflector/NewClassCompiler.cs
```diff
@@ -1730,26 +1728,53 @@ public string ToOperatorName (FunctionDeclaration function)
...
+				var genParts = new StringBuilder ();
+				genParts.Append ('<');
+				for (int i = 0; i < protocolDecl.AssociatedTypes.Count - 1; i++) {
+					if (i > 0)
+						genParts.Append (' ');
+					genParts.Append (',');
+				}
+				genParts.Append ('>');
+				classNameSuffix = genParts.ToString ();
+			}
```
mandel-macaque: I believe you can make all this code a lot simpler, something of the like:
```csharp
var classNameSuffix = $"<{new String (',', protocolDecl.AssociatedTypes.Count - 1)}>"
```
Only diff I can see is that you are going to miss the ' ' and rather than '<, , ,> you are going to get '<,,,>'. Which is not as nice. Another way maybe that gets closer to what you want probably use concat:
```csharp
var classNameSuffix = $"<{string.Concat(Enumerable.Repeat(", ", protocolDecl.AssociatedTypes.Count - 1))}>";
```
But I think I can do better with Enumerable.Repeat and string.Join
```csharp
var classNameSuffix = $"<{string.Join (Enumerable.Repear (" ", protocolDecl.AssociatedTypes.Count - 1), ", "))}
```
The very last one (I have not executed but I believe it does, will return the same as the loop you have, that is, '<, , ,>'.

Also, AFAIK, string.Join, if we know the size and the array of strings (Enumerable.Repeat should do the trick) is more efficient in terms of memory usage than StringBuilder.Append, yet! uses better memory (less copies blah blah) but will go through the given array 2 times, but I need to look at the actual implementation we have.

mandel-macaque: Forgot to mention, that you can even get the if out and go for a unary operator. Chances are that at point someone will hate me for playing ⛳️

SwiftReflector/NewClassCompiler.cs
```diff
@@ -1730,26 +1728,53 @@ public string ToOperatorName (FunctionDeclaration function)
 
 			string ifaceName = InterfaceNameForProtocol (swiftClassName, TypeMapper);
 			string className = CSProxyNameForProtocol (swiftClassName, TypeMapper);
+			string classNameSuffix = "";
```
mandel-macaque:
```suggestion
			string classNameSuffix = String.Empty;
```

SwiftReflector/NewClassCompiler.cs
Therzok: string.Join and string.Concat are (were) not optimized for `IEnumerable<string>` overloads:
https://source.dot.net/#System.Private.CoreLib/String.Manipulation.cs,630a858d9b6f4c5d
https://source.dot.net/#System.Private.CoreLib/String.Manipulation.cs,fa961ce440526a4d

It seems that recent versions are optimized for small string values. Older versions would not benefit from this and would revert to a simple string.Concat.

Therzok: I would create a benchmarkdotnet benchmark under mono to see how it would behave.

stephen-hawley: That could be done, but keep in mind that from the point of view of BtFS, this is a micro optimization. This particular chunk of code will be called relatively rarely compared to other things.

mandel-macaque: @stephen-hawley we (@Therzok  and I) have fun with this little riddles, don't pay too much attention to us ;)

PR #311, opened 2020-02-07
Repaired an incorrect call in the type accessor so that tests run.

stephen-hawley:
The call to check get the type metadata from swift wasn't present.
Added the pinvoke, changed the call.

There was a fragile test that went funny, so I opened an issue on it and ignored it.


SwiftRuntimeLibrary/SwiftCore.cs
```diff
@@ -381,6 +381,9 @@ struct MetadataResponse {
 			return response.Metadata;
 		}
 
+		[DllImport (kXamGlue, EntryPoint = XamGlueConstants.SwiftCore_GetEnumMetadata)]
+		internal static unsafe extern bool GetEnumMetadataByName (byte* stringData, out SwiftMetatype swiftMetatype);
```
rolfbjarne: You can avoid the unsafe code by doing:

```suggestion
		internal static extern bool GetEnumMetadataByName (byte[] stringData, out SwiftMetatype swiftMetatype);
```

type-o-matic/Program.cs
```diff
@@ -172,7 +172,7 @@ static CSFile GenerateCSharpHashTableFromTypes (TypeAggregator types, PlatformNa
 		public static unsafe bool GetSwiftType (Type t, out SwiftMetatype md) {
 			using (var swiftStr = new SwiftString (t.FullName)) {
 				fixed (byte *swiftData = swiftStr.SwiftData) {
-					return GetSwiftType (swiftData, out md);
+					return SwiftCore.GetEnumMetadataByName (swiftData, out md);
```
rolfbjarne: And then this can be simplified to:
```csharp
using (var swiftStr = new SwiftString (t.FullName)) {
 	return SwiftCore.GetEnumMetadataByName (swiftStr.SwiftData, out md);
}
```

stephen-hawley: I put these changes in. Out of curiosity does it make a practical difference?

rolfbjarne: No, at runtime I wouldn't expect much difference, the main advantage is simpler and shorter C# (also in my experience trying to avoid unsafe code whenever there's a good alternative leads to fewer bugs).

PR #312, opened 2020-02-12
Removed compilation errors in C# wrapping, smoke test passes.

stephen-hawley:
Fixed compilation errors in PAT smoke tests.

There were three basic changes:
1. The C# delegate type that goes in the vtable needed to change. The type of the `self` argument was being set to a `SwiftExistentialContainer`, which is correct for a normal protocol, but not here - it should be `IntPtr`.
2. The call to get the matching C# object for the swift object was treating the `self` argument like a protocol/interface instead of a wrapper class.
3. The proxy class name passed into the marshaler was fine for a typical protocol, but in the case of a wrapper, it needs to have a generic type argument.

Smoke test passes and is no longer ignored.

### DIFF
--- a/SwiftReflector/MarshalEngineCSafeSwiftToCSharp.cs
+++ b/SwiftReflector/MarshalEngineCSafeSwiftToCSharp.cs
@@ -272,14 +272,13 @@ namespace SwiftReflector {
 					invoker = new CSIndexExpression (registryCall, false, callParams.ToArray ());
 				}
 			} else {
+				var thisTypeName = hasAssociatedTypes ? csProxyName : thisType.ToString ();
 				if (thisIsInterface && !hasAssociatedTypes) {
-					var thisTypeName = hasAssociatedTypes ? csProxyName : thisType.ToString ();
 					invoker = new CSFunctionCall (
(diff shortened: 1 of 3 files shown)


SwiftReflector/MarshalEngineCSafeSwiftToCSharp.cs
```diff
@@ -272,15 +272,17 @@ public MarshalEngineCSafeSwiftToCSharp (CSUsingPackages use, List<string> identi
...
-				if (thisIsInterface) {
+				if (thisIsInterface && !hasAssociatedTypes) {
+					var thisTypeName = hasAssociatedTypes ? csProxyName : thisType.ToString ();
 					invoker = new CSFunctionCall (
-						$"SwiftObjectRegistry.Registry.InterfaceForExistentialContainer<{thisType.ToString ()}> (self).{methodName}",
+						$"SwiftObjectRegistry.Registry.InterfaceForExistentialContainer<{thisTypeName}> (self).{methodName}",
 						false, callParams.ToArray ());
 
 				} else {
+					var thisTypeName = hasAssociatedTypes ? csProxyName : thisType.ToString ();
```
spouliot: This looks identical to the previous clause. Why not move it outside the `if` condition ?

Issue #313, opened 2020-02-14
MakeUnrooted doesn't reparent its members


stephen-hawley:
`BaseDeclaration.MakeUnrooted` copies members over but doesn't reparent the members in the new class.
This is probably bad.

It can be fixed easily enough, but I don't know how that will affect the rest of the code base, so make sure the tests pass.
Also, if this gets fixed in `MakeUnrooted` be sure to remove the code in OverrideBuilder which is a patch for that.

PR #314, opened 2020-02-14
Added Associated Type support in Get properties

stephen-hawley:
Added Associated Types support in get properties.

This is only a smoke test - the code doesn't get executed (yet).

For the most part, the work was similar to methods:
- flag when a protocol has associated types
- use the correct method in the static receiver
- use the generic renamer to get the associated type names in the generics

One other things that came up - the way that I handle the wrapper for associated types is that I write a synthetic `ClassDeclaration` and copy all the protocol methods and properties into it. That done, it can get passed to `MethodWrapping` to get turned into a set of functions that we can pinvoke into.
The problem I was having was that every member of the class points to the parent, but when I was looking for the a matching wrapper, I wasn't finding it, and sure enough, the wrapper class didn't have any properties in it.
The problem was that when I called `MakeUnrooted` on the synthetic class, its members lost their parentage. I put in a hack patch instead of fixing `MakeUnrooted` because I'm worried about code that might depend on that behavior. Issue for that is [here](https://github.com/xamarin/binding-tools-for-swift/issues/313).

### DIFF
--- a/SwiftReflector/OverrideBuilder.cs
+++ b/SwiftReflector/OverrideBuilder.cs
@@ -255,7 +255,10 @@ namespace SwiftReflector {
 			dtor.ReturnTypeName = "()";
 			decl.Members.Add (dtor);
 
-			return decl.MakeUnrooted () as ClassDeclaration;
+			decl = decl.MakeUnrooted () as ClassDeclaration;
+			foreach (var member in decl.Members)
+				member.Parent = decl;
+			return decl;
 		}
 
 		void CopyConstructors ()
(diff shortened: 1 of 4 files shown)


SwiftReflector/OverrideBuilder.cs
```diff
@@ -837,7 +840,11 @@ SLProperty DefineOverrideProperty (FunctionDeclaration getter, FunctionDeclarati
 				var ifelse = new SLIfElse (condition, ifblock, elseblock);
 
 				if (isProtocol) {
-					getBlock = ifblock;
+					if (getBlock.Count == 0)
+						getBlock = ifblock;
+					else {
```
rolfbjarne: Nitpick: if one part of the if statement has curly braces, the other should have too (or neither).

stephen-hawley: Good to know. Likely this test is going to go away in about 4 PRs so I'm not going to fix it now.

PR #319, opened 2020-02-21
Added set support to PAT indexers

stephen-hawley:
Fixed an NRE
Removed a dead error message
Added a test


SwiftReflector/MarshalEngineCSafeSwiftToCSharp.cs
```diff
@@ -106,8 +106,7 @@ public MarshalEngineCSafeSwiftToCSharp (CSUsingPackages use, List<string> identi
 				} else if (newValueIsGeneric) {
 					var depthIndex = funcDecl.GetGenericDepthAndIndex (swiftNewValue.TypeSpec);
 					var genRef = new CSGenericReferenceType (depthIndex.Item1, depthIndex.Item2);
-					if (genRef == null)
-						throw ErrorHelper.CreateError (ReflectorError.kTypeMapBase + 30, "Inconceivable! The generic type for a parameter in a method was NOT a CSGenericReferenceType!");
```
spouliot: ^ missed that before

PR #320, opened 2020-02-21
Added a test of an associated type as a subscript parameter

stephen-hawley:
Worked first try.
(looks outside to see if it's raining frogs)

### DIFF
--- a/tests/tom-swifty-test/SwiftReflector/ProtocolConformanceTests.cs
+++ b/tests/tom-swifty-test/SwiftReflector/ProtocolConformanceTests.cs
@@ -289,5 +289,22 @@ public protocol Iterator5 {
 			TestRunning.TestAndExecute (swiftCode, callingCode, "OK\n", platform: PlatformName.macOS);
 		}
 
+		[Test]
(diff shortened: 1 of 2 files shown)


tests/tom-swifty-test/SwiftReflector/ProtocolConformanceTests.cs
```diff
@@ -289,5 +289,22 @@ get set
...
+public protocol Iterator6 {
+	associatedtype Elem
+	subscript (index: Elem) -> Elem {
+		get set
+	}
+}
+";
+			var printer = CSFunctionCall.ConsoleWriteLine (CSConstant.Val ("OK"));
+			var callingCode = CSCodeBlock.Create (printer);
+			TestRunning.TestAndExecute (swiftCode, callingCode, "OK\n", platform: PlatformName.macOS);
```
spouliot: that build but does not execute any of swift/binding code right ?
🐸

stephen-hawley: Smoke test, like all the rest.

PR #342, opened 2020-04-07
Make wrapping work for Associated Type reference

stephen-hawley:
Swift allows this syntax for getting to an associated type:
```
public func foo<T: SomePAT>(a: T, b: T.SomeElement) { }
```
The problem is that as it comes in `T.SomeElement` which is intended to be "the associated type `SomeElement` relative to `T` which is a PAT is indistinguishable from "the module `T` which contains some type named `SomeElement`. The latter is more common.

The problem first manifested in code to wrap these functions because it saw `T` as a module and tried to add an `import T`, which of course doesn't exist.

My solution is to add the method `IsProtocolWithAssociatedTypesFullPath` and refactor `HasModule` in terms of that.

Past that, the wrapping code needs to take the original `T.SomeElement` and turn it into a generic reference in terms of the wrapper function, which will become (likely) `T0.SomeElement`.

The unit test for this is currently ignored because it trips up further down the line and at that point this PR would start to get ungainly.

### DIFF
--- a/SwiftReflector/SwiftXmlReflection/BaseDeclaration.cs
+++ b/SwiftReflector/SwiftXmlReflection/BaseDeclaration.cs
@@ -112,6 +112,44 @@ namespace SwiftReflector.SwiftXmlReflection {
 			return null;
 		}
 
+		public bool IsProtocolWithAssociatedTypesFullPath (NamedTypeSpec named, TypeMapper typeMap)
+		{
+			return OwningProtocolFromGenericWithFullPath (named, typeMap) != null;
+		}
+
+		public ProtocolDeclaration OwningProtocolFromGenericWithFullPath (NamedTypeSpec named, TypeMapper typeMap)
+		{
+			if (named.Name.Contains (".")) {
+				var parts = named.Name.Split ('.');
+				if (IsTypeSpecGeneric (parts [0])) {
+					// I make assertions about why things can't happen. Here's why:
+					// If we have func foo<T:SomeProtocol>(a:T b:T.Foo)
+					// it means that if the T part of T.Foo is a generic, then T HAS to be
+					// constrained to a protocol with associated types
+					// If T.Foo is a path to an associated type, then it
+					var depthIndex = GetGenericDepthAndIndex (parts [0]);
+					if (depthIndex.Item1 < 0 || depthIndex.Item2 < 0)
+						return null;
+					var genDecl = GetGeneric (depthIndex.Item1, depthIndex.Item2);
+					if (genDecl.Constraints.Count != 1) // pretty sure this can't ever happen
+						return null;
+					var inh = genDecl.Constraints [0] as InheritanceConstraint;
+					if (inh == null) // pretty sure this also can't ever happen
+						return null;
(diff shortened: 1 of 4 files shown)


SwiftReflector/SwiftXmlReflection/BaseDeclaration.cs
```diff
@@ -112,6 +112,44 @@ static ProtocolDeclaration ThisOrParentProtocol (BaseDeclaration self)
...
 
+		public bool IsProtocolWithAssociatedTypesFullPath (NamedTypeSpec named, TypeMapper typeMap)
+		{
+			return OwningProtocolFromGenericWithFullPath (named, typeMap) != null;
+		}
+
+		public ProtocolDeclaration OwningProtocolFromGenericWithFullPath (NamedTypeSpec named, TypeMapper typeMap)
+		{
+			if (named.Name.Contains (".")) {
+				var parts = named.Name.Split ('.');
```
spouliot, line 123: why both `Contains` and `Split` ?
if you're not worried about perf you could do `Split` only (if more than one part then `.` was present)
if you think it's an hot path then an `IndexOf` and a `Substring`) looks better: no array allocation and you're only concerned with the first part

SwiftReflector/SwiftXmlReflection/TypeSpec.cs
```diff
@@ -353,7 +356,7 @@ protected override bool LLEquals (TypeSpec other, bool partialNameMatch)
 		}
 		public string NameWithoutModule {
 			get {
-				return HasModule ? Name.Substring (Name.IndexOf ('.') + 1) : Name;
+				return Name.Contains ('.') ? Name.Substring (Name.IndexOf ('.') + 1) : Name;
```
spouliot, line 359: minor: there's (sadly) no overload for `Constains` that accept characters (it will become a string), `IndexOf` would work on a `char`

SwiftReflector/TypeMapping/TypeSpecToSLType.cs
```diff
@@ -58,15 +58,23 @@ public SLType MapType (BaseDeclaration declContext, SLImportModules modules, Typ
...
 			} else if (spec.ContainsGenericParameters) {
 				retval = new SLBoundGenericType (spec.NameWithoutModule, spec.GenericParameters.Select (p => MapType (declContext, modules, p, false)));
 			} else {
-				retval = new SLSimpleType (spec.Name.NameWithoutModule ());
+				if (declContext.IsProtocolWithAssociatedTypesFullPath (spec, parent)) {
+					// for T.AssocType
+					var genPart = spec.Module;
+					var depthIndex = declContext.GetGenericDepthAndIndex (genPart);
+					var newGenPart = new SLGenericReferenceType (depthIndex.Item1, depthIndex.Item2);
+					retval = new SLSimpleType ($"{newGenPart.ToString ()}.{spec.NameWithoutModule}");
```
spouliot, line 74: minor: `.ToString` is not needed, it will be called automatically

SwiftReflector/SwiftXmlReflection/TypeSpec.cs
stephen-hawley: 👍

Issue #347, opened 2020-04-16
Associated type return value not being marshaled properly


stephen-hawley:
```
public protocol Simplest4 {
        associatedtype Item
        subscript (index: Int) -> Item {
                get
        }
}
public func doGetIt<T:Simplest4> (a: T, i: Int) -> T.Item {
        return a[i]
}
```
`doGetIt` should be wrapped to take a pointer to a return value and write into it.

stephen-hawley:
Fixed [here](https://github.com/xamarin/binding-tools-for-swift/pull/348).

PR #348, opened 2020-04-16
Fix for marshaling associated type paths in return values

stephen-hawley:
This fixes issue [347](https://github.com/xamarin/binding-tools-for-swift/issues/347) when marshaling a return value that is an associated type path

Marshaling was wrong at a number of levels:
- the function `MustBeInOut` in MethodWrapping needed a case for associated type paths
- TypeMapper had a number of functions that needed cases for associated type paths
- finally MarshalEngine needed to handle associated type paths as pointers appropriately

The code for marshaling from C# to swift and back is very similar to the code for handling a pointer to a generic type, but there are enough differences that it would have taken away from the readability of unified code to handle both.

### DIFF
--- a/SwiftReflector/MarshalEngine.cs
+++ b/SwiftReflector/MarshalEngine.cs
@@ -219,6 +219,7 @@ namespace SwiftReflector {
 			var returnIsNonTrivialTuple = hasReturn && swiftReturnType is TupleTypeSpec && ((TupleTypeSpec)swiftReturnType).Elements.Count > 1;
 			var returnIsClosure = hasReturn && swiftReturnType is ClosureTypeSpec;
 			var returnIsGeneric = hasReturn && typeContext.IsTypeSpecGeneric (swiftReturnType) && !returnIsClosure;
+			var returnIsAssocPath = hasReturn && typeContext.IsProtocolWithAssociatedTypesFullPath (swiftReturnType as NamedTypeSpec, typeMapper);
 			var returnIsNonScalarStruct = hasReturn && !returnIsScalar && returnEntity != null &&
 				(returnEntity.EntityType == EntityType.Struct || returnEntity.EntityType == EntityType.Enum);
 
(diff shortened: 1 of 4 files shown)


SwiftReflector/MarshalEngine.cs
```diff
@@ -944,6 +950,44 @@ CSBaseExpression MarshalProtocolListAsPointer (BaseDeclaration typeContext, CSPa
...
+			preMarshalCode.Add (retvalBufferPtrDecl);
+
+			if (!isReturnValue) {
+				preMarshalCode.Add (CSFunctionCall.FunctionCallLine ("StructMarshal.Marshaler.ToSwift", false,
+					p.CSType.Typeof (), p.Name, bufferPtrIdent));
+			}
+			var toNetCall = new CSFunctionCall (String.Format ("StructMarshal.Marshaler.ToNet<{0}>", p.CSType.ToString ()),
+				  false, bufferPtrIdent, CSConstant.Val (true));
+			postMarshalCode.Add (new CSLine (toNetCall));
+			return bufferPtrIdent;
```
whitneyschmidt, line 988: If I am understanding this code (doubtful...), the return value is a `CSIdentifier` that stores a unique string to identify the C# function call that is created here.

stephen-hawley: You're close.
Most every function of the form `MarshalSomeType` returns a `CSBaseExpression` which represents an expression that gets passed into swift to represent the parameter `p`. In some cases, it's easy. If the type of `p` is a scalar, I can return `p` as an identifier, since that goes into swift with no changes.
If the the of `p` is `ISwiftObject` then I pass in `p.SwiftObject` which is the handle to underlying swift object.
In the case of an associated type path (which is a generic), there is pre-marshaling code which in a typical case:
1 - stack allocates a buffer the size of the swift type the `p` represents
2 - does runtime marshaling to put the appropriate value of that buffer
3 - makes an `IntPtr` from that buffer which will get passed in to the swift function
Upon return from calling the swift function, nothing else happens.
Now because I'm super lazy, if the return value of a function won't necessarily match our register return calling conventions, I do a trick where I inject the return value as a fake C# parameter with a flag saying "this is a return value". It acts almost the same as the parameter case but with two differences:
1 - it doesn't do runtime marshaling to copy into the buffer because there is no value
2 - it adds post marshaling code to put unmarshal the return value from the buffer into something C#-ish

whitneyschmidt: @stephen-hawley Thanks for the explanation! The return value special case is particularly mind-bendy. 👍

PR #351, opened 2020-04-21
Added predicate for dynamic self

stephen-hawley:
We need the ability to detect if a protocol contains the type `Self` which is the dynamic self type.

To do this, I added code in `ProtocolDeclaration` to search through all the members for it as a type. This ends up being recursive to search through tuples, closures, etc.

In the process, I discovered an issue with the reflector in that if a protocol comes through as a parameter type, it always is listed as `Self` which is incorrect - it should be the full name of the protocol. The problem is that if the type really was `Self` it needs to stay self. So how do distinguish between the two cases?

In the reflector, I added a call to `existentialTypeSupported` which returns `true` if and only if a type can be represented by an existential type. If a type has `Self` or an associated type, it can't be an existential type, so if `existentialTypeSupported` returns false means that the `Self` is appropriate.

Tests as per usual.

### DIFF
--- a/SwiftReflector/SwiftXmlReflection/ProtocolDeclaration.cs
+++ b/SwiftReflector/SwiftXmlReflection/ProtocolDeclaration.cs
@@ -51,7 +51,7 @@ namespace SwiftReflector.SwiftXmlReflection {
 		}
 
 		public bool HasDynamicSelf {
-			// you could cache this, but this type is not mutable, so that would be bad
+			// you shouldn't cache this. This type is mutable, so that would be bad
 			get => SearchForDynamicSelf ();
 		}
 
--- a/SwiftReflector/SwiftXmlReflection/ProtocolDeclaration.cs
+++ b/SwiftReflector/SwiftXmlReflection/ProtocolDeclaration.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Xml.Linq;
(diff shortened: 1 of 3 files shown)


SwiftReflector/SwiftXmlReflection/ProtocolDeclaration.cs
```diff
@@ -48,6 +49,72 @@ public AssociatedTypeDeclaration AssociatedTypeNamed (string name)
...
+			var types = funcDecl.ParameterLists.Last ().Select (p => p.TypeSpec).ToList ();
+			if (!TypeSpec.IsNullOrEmptyTuple (funcDecl.ReturnTypeSpec))
+				types.Add (funcDecl.ReturnTypeSpec);
+			return SearchForDynamicSelf (types);
+		}
+
+		bool SearchForDynamicSelf (PropertyDeclaration propDecl)
+		{
+			var types = new List<TypeSpec> ();
+			types.Add (propDecl.TypeSpec);
```
spouliot, line 86: that extraneous list (already in a loop, inside recursion) can be removed if you implement
`bool SearchForDynamicSelf (TypeSpec type)`
and then make
`bool SearchForDynamicSelf (List<TypeSpec> types)` use it

SwiftReflector/SwiftXmlReflection/ProtocolDeclaration.cs
```diff
@@ -48,6 +49,72 @@ public AssociatedTypeDeclaration AssociatedTypeNamed (string name)
...
+					if (SearchForDynamicSelf (ns))
+						return true;
+				} else if (type is TupleTypeSpec tuple) {
+					if (SearchForDynamicSelf (tuple.Elements))
+						return true;
+				} else if (type is ClosureTypeSpec closure) {
+					var moreTypes = new List<TypeSpec> ();
+					moreTypes.Add (closure.Arguments);
+					if (!TypeSpec.IsNullOrEmptyTuple (closure.ReturnType))
+						moreTypes.Add (closure.ReturnType);
```
spouliot, line 103: the `else` clause could also avoid the `moreTypes` list creation (same code as above)

SwiftReflector/SwiftXmlReflection/ProtocolDeclaration.cs
```diff
@@ -48,6 +49,78 @@ public AssociatedTypeDeclaration AssociatedTypeNamed (string name)
 		{
 			return AssociatedTypes.FirstOrDefault (at => at.Name == name);
 		}
+
+		public bool HasDynamicSelf {
+			// you could cache this, but this type is not mutable, so that would be bad
```
rolfbjarne, line 54: Is this comment correct? Sounds like caching something would be bad if the type were mutable, not if it weren't.

stephen-hawley: Good catch.

PR #357, opened 2020-05-12
Handle dynamic self in returns

stephen-hawley:
This is a smoke test - it writes swift wrappers and C# bindings and both compile correctly.
Does it run? 🤷 That's for another day.

How does this work? ALL THE SPECIAL CASES.
- made a change to the code that matches wrappers to original functions to understand about self. This gets done by substituting the parent protocol type for Self.
- marshaling code needed to know about self. This code will fail with Self in arguments. That's OK, that's more special cases for another day
- made a new type of `NetTypeBundle` to understand dynamic self (and wishing that C# had discriminated unions)
- made the type mapper understand dynamic sel

Smoke test now passes.

### DIFF
--- a/SwiftReflector/MarshalEngine.cs
+++ b/SwiftReflector/MarshalEngine.cs
@@ -222,13 +222,15 @@ namespace SwiftReflector {
 			var returnIsAssocPath = hasReturn && typeContext.IsProtocolWithAssociatedTypesFullPath (swiftReturnType as NamedTypeSpec, typeMapper);
 			var returnIsNonScalarStruct = hasReturn && !returnIsScalar && returnEntity != null &&
 				(returnEntity.EntityType == EntityType.Struct || returnEntity.EntityType == EntityType.Enum);
+			var returnIsSelf = hasReturn && swiftReturnType.IsDynamicSelf;
 
 			var retSimple = returnType as CSSimpleType;
 			var returnIsInterfaceFromProtocol =
 				hasReturn && returnEntity != null && returnEntity.EntityType == EntityType.Protocol && retSimple != null;
 			var returnIsProtocolList = hasReturn && swiftReturnType is ProtocolListTypeSpec;
 			var returnIsObjCProtocol = hasReturn && returnEntity != null && returnEntity.IsObjCProtocol;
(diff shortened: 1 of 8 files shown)


SwiftReflector/MarshalEngine.cs
```diff
@@ -222,13 +222,15 @@ CSLine CreateNominalCall (CSType returnType, CSIdentifier id, CSType objectType,
...
+			var returnIsSelf = hasReturn && swiftReturnType.IsDynamicSelf;
 
 			var retSimple = returnType as CSSimpleType;
 			var returnIsInterfaceFromProtocol =
 				hasReturn && returnEntity != null && returnEntity.EntityType == EntityType.Protocol && retSimple != null;
 			var returnIsProtocolList = hasReturn && swiftReturnType is ProtocolListTypeSpec;
 			var returnIsObjCProtocol = hasReturn && returnEntity != null && returnEntity.IsObjCProtocol;
-			var returnNeedsPostProcessing = (hasReturn && (returnIsClass || returnIsProtocolList || returnIsInterfaceFromProtocol || returnIsNonTrivialTuple || returnIsGeneric || returnIsNonScalarStruct || returnIsAssocPath)) || originalThrows;
+			var returnNeedsPostProcessing = (hasReturn && (returnIsClass || returnIsProtocolList || returnIsInterfaceFromProtocol || returnIsNonTrivialTuple ||
+				returnIsGeneric || returnIsNonScalarStruct || returnIsAssocPath || returnIsSelf)) || originalThrows;
```
dalexsoto, line 233: You are making the case here for me to get a 49" ultra wide monitor 👍

SwiftReflector/SwiftXmlReflection/ParameterItem.cs
```diff
@@ -165,6 +169,15 @@ public static bool AreEqualIgnoreNames (IList<ParameterItem> pl1, IList<Paramete
 			return true;
 		}
 
+		static ParameterItem SubstituteSelfFromParent (FunctionDeclaration func, ParameterItem p)
+		{
+			if (func.Parent == null || !p.TypeSpec.HasDynamicSelf)
+				return p;
+			p = new ParameterItem (p);
+			p.TypeSpec = p.TypeSpec.ReplaceName ("Self", func.Parent.ToFullyQualifiedNameWithGenerics ());
```
spouliot, line 177: you just re-assigned `p` so is `p.TypeSpec` what you want (the new one vs the one given as an argument) ?

stephen-hawley: Yes. We run it through a copy constructor to not alter the input and then side-effect the result.

PR #360, opened 2020-05-21
Execute dynamic self in unit test

stephen-hawley:
The goal this PR was to execute code in C# that implements an interface using dynamic self only in a return.

There were three issues that came to light from this test:

1. The C# code for the top level function did not compile because of an incorrect function signature. The generic parameter for `TSelf` needed to get added into the function signature. Fortunately, there is a code path to do that in `TypeMapper.cs` for generics with constraints, so the code there just detects this special case and injects `TSelf` and its constraint.
2. There was a runtime error when making the proxy. Necessarily, in `SwiftProtocolAttribute`, the type of the proxy is unbound, eg, `FooProxy<,,,>`, but the proxy that gets created needs to have concrete types, so find the interface via reflection and extract the types from it and inject them into the proxy type and use this to make the proxy.
3. The vtable setter pinvoke and the corresponding C# implementation was incorrectly trying to pass metadata types into swift. This isn't correct behavior, so the code the generates the vtable setter and the pinvoke declaration needed to only add `SwiftMetatype` arguments if it was a real generic type, as opposed to this case, which is an `EveryProtocol` in swift which has no generics.

### DIFF
--- a/SwiftRuntimeLibrary/SwiftMarshal/SwiftProtocolTypeAttribute.cs
+++ b/SwiftRuntimeLibrary/SwiftMarshal/SwiftProtocolTypeAttribute.cs
@@ -52,6 +52,7 @@ namespace SwiftRuntimeLibrary.SwiftMarshal {
 		public static T MakeProxy<T> (Type proxyType, T interfaceImpl)
 		{
 			var interfaceType = typeof (T);
+			proxyType = RarefiedProxyType (interfaceImpl, ref interfaceType, proxyType);
 			var ci = proxyType.GetConstructor (new Type [] { interfaceType });
 			if (ci == null)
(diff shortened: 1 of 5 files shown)


SwiftRuntimeLibrary/SwiftMarshal/SwiftProtocolTypeAttribute.cs
```diff
@@ -69,6 +71,31 @@ public static BaseProxy MakeProxy (Type interfaceType, object interfaceImpl, Eve
 			return (BaseProxy)ci.Invoke (new object [] { interfaceImpl, protocol });
 		}
 
+		static Type RarefiedProxyType (object interfaceImpl, ref Type interfaceType, Type proxyType)
```
spouliot, line 74: nice name :)

PR #367, opened 2020-07-14
Added code to support access to implicit arguments.

stephen-hawley:
We have a challenge in working with swift that there are times when swift has implicit arguments to functions. Here's an example:
```
public protocol AProto {
    func x ()
}
public func doX<T: AProto> (x: T) {
    x.x()
}
```

In `doX` there appears to be only 1 argument, but in reality there are three: `x`, the type metadata for `T` and the protocol witness table for `T` with respect to `AProto`.

There are times when we will need to access the protocol witness table in order to pass that to C#, but because it's implicit, we're out of luck doing that in swift. Or are we?

If we write a function like this in swift:
```
@inline(never)
public func protocolWitness<T: AProto>(of: T) -> UnsafeRawPointer
{
    return RegisterAccess.swiftAsmArg2()!
}
```
We can get that argument. How do we do this? We do the terrible things that C lets you do. In this case, we write a function `swiftAsmArg2()` which looks like it takes no arguments and returns `const void *`. In reality, what it does is access the caller's arguments instead. This happens in two ways. The first way is that if the CPU ABI dictates using register passing, we write code to stuff the register argument into the return register. If the CPU supports stack passed arguments, we reach past our own stack frame to get the arguments.

How do we know that this is safe-ish? If we embed a call to the C inside a swift function like `protocolWitnessOf` above which has no local variables, then the arguments registers and stack is untouched. This is true in debug and it is true in release.

How does swift import the C code? We create a file called `module.map` that declares that all the global declarations in a given header file will end up in the module `RegisterAccess`. This means that to compile against our code, we need to ship the `module.map` file as well as `registeraccess.h`.

There are implementations for armv7*, i386, x64, and arm64.

There is a test - stunt testing!
I define a protocol, implement it on a class, then write a `protocolWitness` function that returns a given protocol witness table for an implementor of a protocol. I call this with the C# binding to the implementation and compare the return value to `StructMarshal.Marshaler.ProtocolWitnessof`. The value that comes back matches in a correct case.

### DIFF
--- a/swiftglue/Makefile
+++ b/swiftglue/Makefile
@@ -25,6 +25,8 @@ bindingmetadata.macos.swift:
 
 generate-swift-bindings: bindingmetadata.iphone.swift bindingmetadata.macos.swift
 
+CFILES:=$(wildcard *.c)
+
 #
 # Define a template that can compile the swift glue code for each architecture
 #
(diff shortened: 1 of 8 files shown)


swiftglue/Makefile
```diff
@@ -37,17 +37,27 @@ $(3)_SDK:=$$(shell xcrun --sdk $(2) --show-sdk-path)
 
 ## debug
 $(3)_DEBUG_OUTPUTDIR:=$(BINDIR)/Debug/$(1)/$$($(3)_ARCH)
-$$($(3)_DEBUG_OUTPUTDIR)/$(MODULENAME): Makefile $(wildcard *.swift)
+$(3)_DEBUG_COUTPUTDIR:=$(BINDIR)/Debug/$(1)/$$($(3)_ARCH)/ofiles
+$$($(3)_DEBUG_COUTPUTDIR)/%.o : %.c
+	mkdir -p $$($(3)_DEBUG_COUTPUTDIR)
+	clang -x c -arch $$($(3)_ARCH) -std=gnu11 -O0 -fasm-blocks -c $$< -o $$($(3)_DEBUG_COUTPUTDIR)/$$<.o
+
+$$($(3)_DEBUG_OUTPUTDIR)/$(MODULENAME): Makefile $(wildcard *.swift) $$($(3)_DEBUG_COUTPUTDIR)/*.o
```
rolfbjarne, line 45: If you start with a clean directory, the *.o files won't be there, which means they won't be added to the dependencies, which means they won't be included in the build. I think something like this should work:

```csharp
# this should go outside the SwiftLibraryTemplate template
CFILES:=$(wildcard *.c)

# this should go inside the SwiftLibraryTemplate template
DEBUG_OBJFILES:=$$(foreach cfile,$(CFILES),$$($(3)_DEBUG_OUTPUTDIR)/$$(cfile:.c=.o))

$$($(3)_DEBUG_OUTPUTDIR)/$(MODULENAME): Makefile $(wildcard *.swift) $$(DEBUG_OBJFILES)
	...
```

swiftglue/Makefile
```diff
@@ -37,17 +37,27 @@ $(3)_SDK:=$$(shell xcrun --sdk $(2) --show-sdk-path)
 
 ## debug
 $(3)_DEBUG_OUTPUTDIR:=$(BINDIR)/Debug/$(1)/$$($(3)_ARCH)
-$$($(3)_DEBUG_OUTPUTDIR)/$(MODULENAME): Makefile $(wildcard *.swift)
+$(3)_DEBUG_COUTPUTDIR:=$(BINDIR)/Debug/$(1)/$$($(3)_ARCH)/ofiles
+$$($(3)_DEBUG_COUTPUTDIR)/%.o : %.c
+	mkdir -p $$($(3)_DEBUG_COUTPUTDIR)
```
rolfbjarne, line 42: This makes the output nicer/quiet:

```suggestion
	$$(Q) mkdir -p $$($(3)_DEBUG_COUTPUTDIR)
```

swiftglue/Makefile
```diff
@@ -37,17 +37,27 @@ $(3)_SDK:=$$(shell xcrun --sdk $(2) --show-sdk-path)
...
+	clang -x c -arch $$($(3)_ARCH) -std=gnu11 -O0 -fasm-blocks -c $$< -o $$($(3)_DEBUG_COUTPUTDIR)/$$<.o
```
rolfbjarne, line 43: Same, nicer output:

```suggestion
	$$(call Q_2,CLANG [$(1)/$$($(3)_ARCH)/Debug/$$(@F)]) clang -x c -arch $$($(3)_ARCH) -std=gnu11 -O0 -fasm-blocks -c $$< -o $$($(3)_DEBUG_COUTPUTDIR)/$$<.o
```

swiftglue/Makefile
```diff
@@ -37,17 +37,27 @@ $(3)_SDK:=$$(shell xcrun --sdk $(2) --show-sdk-path)
...
 	$$(Q) install_name_tool -delete_rpath $(realpath $(UNPACKAGED_SWIFTLIB)/$(2)) $$@.tmp
 	$$(Q) mv $$@.tmp $$@
 
 ## release
 $(3)_RELEASE_OUTPUTDIR=$(BINDIR)/Release/$(1)/$$($(3)_ARCH)
-$$($(3)_RELEASE_OUTPUTDIR)/$(MODULENAME): Makefile $(wildcard *.swift)
+$(3)_RELEASE_COUTPUTDIR:=$(BINDIR)/Release/$(1)/$$($(3)_ARCH)/ofiles
+$$($(3)_RELEASE_COUTPUTDIR)/%.o : %.c
+	mkdir -p $$($(3)_RELEASE_COUTPUTDIR)
+	clang -x c -arch $$($(3)_ARCH) -DCPU_ARCH=$$($(3)_ARCH) -std=gnu11 -Os -fasm-blocks -c $$< -o $$($(3)_RELEASE_COUTPUTDIR)/$$<.o
```
rolfbjarne, line 56: Same:

```suggestion
	$$(Q) mkdir -p $$($(3)_RELEASE_COUTPUTDIR)
	$$(call Q_2,CLANG [$(1)/$$($(3)_ARCH)/Release/$$(@F)]) clang -x c -arch $$($(3)_ARCH) -DCPU_ARCH=$$($(3)_ARCH) -std=gnu11 -Os -fasm-blocks -c $$< -o $$($(3)_RELEASE_COUTPUTDIR)/$$<.o
```

swiftglue/Makefile
```diff
@@ -37,17 +37,27 @@ $(3)_SDK:=$$(shell xcrun --sdk $(2) --show-sdk-path)
...
+
+$$($(3)_DEBUG_OUTPUTDIR)/$(MODULENAME): Makefile $(wildcard *.swift) $$($(3)_DEBUG_COUTPUTDIR)/*.o
 	$(Q) mkdir -p $$(dir $$@)
-	$$(call Q_2,SWIFTC [$(1)/$$($(3)_ARCH)/Debug])   $(UNPACKAGED_SWIFTC) -g -sdk $$($(3)_SDK) -target $(3) -emit-module -emit-library -Xlinker -rpath -Xlinker /usr/lib/swift -Xlinker -rpath -Xlinker @executable_path/Frameworks -Xlinker -rpath -Xlinker @loader_path/Frameworks -Xlinker -rpath -Xlinker @executable_path -Xlinker -final_output -Xlinker $(MODULENAME) -Xlinker -install_name -Xlinker @rpath/$(MODULENAME).framework/$(MODULENAME) -module-name $(MODULENAME) -o $$@.tmp *.swift
+	$$(call Q_2,SWIFTC [$(1)/$$($(3)_ARCH)/Debug])   $(UNPACKAGED_SWIFTC) -g -sdk $$($(3)_SDK) -target $(3) -emit-module -emit-library -Xlinker -rpath -Xlinker /usr/lib/swift -Xlinker -rpath -Xlinker @executable_path/Frameworks -Xlinker -rpath -Xlinker @loader_path/Frameworks -Xlinker -rpath -Xlinker @executable_path -Xlinker -final_output -Xlinker $(MODULENAME) -Xlinker -install_name -Xlinker @rpath/$(MODULENAME).framework/$(MODULENAME) -I . -module-name $(MODULENAME) -o $$@.tmp *.swift $$($(3)_DEBUG_COUTPUTDIR)/*.o
```
rolfbjarne, line 47: This is not needed, but I think it makes sense now (using DEBUG_OBJFILES instead of a glob):

```suggestion
	$$(call Q_2,SWIFTC [$(1)/$$($(3)_ARCH)/Debug])   $(UNPACKAGED_SWIFTC) -g -sdk $$($(3)_SDK) -target $(3) -emit-module -emit-library -Xlinker -rpath -Xlinker /usr/lib/swift -Xlinker -rpath -Xlinker @executable_path/Frameworks -Xlinker -rpath -Xlinker @loader_path/Frameworks -Xlinker -rpath -Xlinker @executable_path -Xlinker -final_output -Xlinker $(MODULENAME) -Xlinker -install_name -Xlinker @rpath/$(MODULENAME).framework/$(MODULENAME) -I . -module-name $(MODULENAME) -o $$@.tmp *.swift $$(DEBUG_OBJFILES)
```

swiftglue/Makefile
```diff
@@ -37,17 +37,27 @@ $(3)_SDK:=$$(shell xcrun --sdk $(2) --show-sdk-path)
...
 
 ## release
 $(3)_RELEASE_OUTPUTDIR=$(BINDIR)/Release/$(1)/$$($(3)_ARCH)
-$$($(3)_RELEASE_OUTPUTDIR)/$(MODULENAME): Makefile $(wildcard *.swift)
+$(3)_RELEASE_COUTPUTDIR:=$(BINDIR)/Release/$(1)/$$($(3)_ARCH)/ofiles
+$$($(3)_RELEASE_COUTPUTDIR)/%.o : %.c
+	mkdir -p $$($(3)_RELEASE_COUTPUTDIR)
+	clang -x c -arch $$($(3)_ARCH) -DCPU_ARCH=$$($(3)_ARCH) -std=gnu11 -Os -fasm-blocks -c $$< -o $$($(3)_RELEASE_COUTPUTDIR)/$$<.o
+
+$$($(3)_RELEASE_OUTPUTDIR)/$(MODULENAME): Makefile $(wildcard *.swift) $$($(3)_RELEASE_COUTPUTDIR)/*.o
```
rolfbjarne, line 58: And the same change here with regards to the object files.

swiftglue/registeraccess.c
```diff
@@ -0,0 +1,100 @@
...
+#elif __arm
+	__asm("mov r0, r0");
+#elif i386
+	asm {
+		mov eax, dword ptr [ebp + 12]
+	}
+#else
+#error("Unknown CPU type for register access");
+#endif
+}
```
rolfbjarne, line 32: Wouldn't this work just as well?

```c
void *swiftArg0 (void *p0)
{
    return p0:
}

void *swiftArg1 (void *p0, void *p1)
{
    return p1;
}
...
```

And it's bitcode-compatible!

See: https://gist.github.com/rolfbjarne/6181be7133c129670196c43619ac744d

stephen-hawley: Oh this is hilarious - lie about the signature in the export vis the implementation! Have verified that it works.

swiftglue/registeraccess.c
```diff
@@ -0,0 +1,54 @@
...
+
+void swiftSelfArg ()
+{
+#if __x86_64
+	__asm("mov %r13, %rax");
+#elif __arm64
+	__asm("mov x0, x20");
+#elif __arm
+	__asm("mov r0, r0");
```
rolfbjarne, line 46: I think C code to return the first argument would work for arm (and i386) as well, seems like that's where the self argument is passed?

swiftglue/Makefile
stephen-hawley: These sets of changes aren't quite right - I get `make: *** No rule to make target `bin/Debug/mac/x86_64/registeraccess.o', needed by `bin/Debug/mac/x86_64/XamGlue'.  Stop.`

swiftglue/registeraccess.c
stephen-hawley: The self argument depends on the architecture. In x64, its R13. In x86 and arm7 it's the first argument. In arm64 it's x20.

stephen-hawley: I have no idea what bitcode does.

swiftglue/Makefile
rolfbjarne: Something like this seems to work: https://gist.github.com/rolfbjarne/a52b046d93957e42d884651a73331e16

PR #370, opened 2020-07-23
[DevOps] Add a tap for the project.

mandel-macaque:
Add a tap that contains different casks to be used with the project so
that we do not need to do anything special regarding cmake and other
deps.

### DIFF
--- a/Casks/README.md
+++ b/Casks/README.md
@@ -13,14 +13,14 @@ It is important to use the git url else brew will default to https and will ask 
 has been added you can use brew as you normally do:
 
 ```bash
-brew cask install cmake-btfw
+brew cask install cmake-btfs
 ```
 
 # Casks
 
 The following casks are available:
 
-## cmake-btfw
+## cmake-btfs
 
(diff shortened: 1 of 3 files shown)


Casks/cmake-btfw.rb
```diff
@@ -0,0 +1,18 @@
+cask 'cmake-btfw' do
```
stephen-hawley: Probably need to include the copyright:
```
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.
```

Casks/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+```bash
+brew tap xamarin/binding-tools-for-swift git@github.com:xamarin/binding-tools-for-swift.git
+```
+
+It is important to use the git url else brew will default to https and will ask for the github credentials. Once the tap
+has been added you can use brew as you normally do:
+
+```bash
+brew cask install cmake-btfw
```
spouliot: `btfs` ?

Issue #142, opened 2018-04-23
[SoM] Non trivial properties on trivial enums have incorrect signature.


stephen-hawley:
Given this swift enum:
```
public enum Rocks {
    case igneous, sedimentary, metamorphic
    public var Rocks: String { return "Pile Of Rocks" }
}
```
Should generate an extension method:
```
public static SwiftString Rocks(this Rocks) { /* ... */ }
```
But it has the wrong signature.

stephen-hawley:
Fixed [here](https://github.com/xamarin/binding-tools-for-swift/pull/377).

PR #377, opened 2020-08-11
Fixed signature of trivial enum property

stephen-hawley:
Corrected the signature of trivial enum property fixing [issue 142](https://github.com/xamarin/binding-tools-for-swift/issues/142)

This is an exploration in how you can make life easier for your future self. In the issue above, past Steve went to the trouble of including the swift code that triggered the issue and present Steve is very happy about that.

Now, here's what was going on. Given this swift code:
```
public enum Rocks {
    case igneous, sedimentary, metamorphic
    public var Rocks: String { return ""Pile Of Rocks"" }
}
```

BTfS was generating an extension method on the C# enum `Rocks` that had this signature:
```
public static void Rocks (this SwiftString retval, Rocks this0)
{
}
```
which is totally wrong. The reason is that we have two functions representing the method: The actual swift getter (which we can't call) and a wrapper that we can call. We were using only the getter wrapper before which has this following signature:
```
func Rocks (retval: UnsafeMutablePointer<String>, this: Rocks)
```
What we really want to do is use the getter signature (`func Rocks () -> String`) for generating the signature of the function. This gives us this C# signature:
```
public static SwiftString Rocks ()
{
}
```
Which is close, but we need the `this` parameter. We get that by also generating the signature for the wrapper function to get the type of the `this` parameter.  I could type map the swift type to the C#, but there can be issues and special cases in the mapping and `CompileMethod` already handles those. The last parameter is the instance, so we grab that and inject it into the the parameter list.

Test as per usual.

### DIFF
--- a/SwiftReflector/NewClassCompiler.cs
+++ b/SwiftReflector/NewClassCompiler.cs
@@ -4937,9 +4937,12 @@ namespace SwiftReflector {
 			if (prop.TLFGetter != null) {
 				var marshaler = new MarshalEngine (use, useLocals, TypeMapper, wrapper.Module.SwiftCompilerVersion);
 
-				var getter = TLFCompiler.CompileMethod (getterWrapperFunc, use, null, null, getterName, false, false, true);
-				var thisParm = getter.Parameters [0];
-				getter.Parameters [0] = new CSParameter (thisParm.CSType, thisParm.Name, CSParameterKind.This);
+				var wrapperGetter = TLFCompiler.CompileMethod (getterWrapperFunc, use, null, null, getterName, false, false, true);
+
+				var getter = TLFCompiler.CompileMethod (propDecl.GetGetter (), use, null, null, getterName, false, false, true);
+
+				var thisParm = wrapperGetter.Parameters.Last ();
+				getter.Parameters.Add (new CSParameter (thisParm.CSType, thisParm.Name, CSParameterKind.This));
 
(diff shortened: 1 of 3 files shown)


SwiftReflector/NewClassCompiler.cs
```diff
@@ -4937,9 +4937,12 @@ static bool TypesMatch (FunctionDeclaration parentDecl, NamedTypeSpec spec, Swif
 			if (prop.TLFGetter != null) {
 				var marshaler = new MarshalEngine (use, useLocals, TypeMapper, wrapper.Module.SwiftCompilerVersion);
 
-				var getter = TLFCompiler.CompileMethod (getterWrapperFunc, use, null, null, getterName, false, false, true);
-				var thisParm = getter.Parameters [0];
-				getter.Parameters [0] = new CSParameter (thisParm.CSType, thisParm.Name, CSParameterKind.This);
+				var wrapperGetter = TLFCompiler.CompileMethod (getterWrapperFunc, use, null, null, getterName, false, false, true);
```
spouliot, line 4940: Side note: it's getting hard (the diff) to review with so many `null` and boolean arguments.
Maybe you can start to use named arguments ?
Not sure it would help you (you know the calls) but reviewers and possible contributors would have an easier time.

stephen-hawley: 👍 https://github.com/xamarin/binding-tools-for-swift/issues/378

Issue #141, opened 2018-04-24
[SoM] enums are not showing constructors


stephen-hawley:
```
public enum TheForce {
    case do, doNot
    public init(yoda: Bool) {
        self = yoda ? .do : .doNot // there is no try
    }
}
```
Needs to have factory methods for constructors in the sidecar class (that holds extensions right now).

Also ensure that optional constructors work too:
```
		[Test]
		public void TestSimpleEnumOptionalCtor ()
		{
			string swiftCode =
				"public enum FirstOptEnumCtor {\n" +
				"    case good, bad\n" +
				"    public init?(fail: Bool) {\n" +
				"        if fail {\n" +
				"            return nil\n" +
				"        }\n" +
				"        self = FirstOptEnumCtor.good\n" +
				"    }\n" +
				"}\n";

			var varDecl = CSVariableDeclaration.VarLine (new CSSimpleType ("SwiftOptional<FirstOptEnumCtor>"), "optVal",
			                                             new CSFunctionCall ("FirstOptEnumCtor.FirstOptEnumCtorOptional", false, CSConstant.Val (true)));
			var printIt = CSFunctionCall.FunctionCallLine ("Console.WriteLine", false, new CSIdentifier ("optVal").Dot (new CSIdentifier ("HasValue")));
			var callingCode = new CodeElementCollection<ICodeElement> () {
				varDecl, printIt
			};
			TestRunning.TestAndExecute (swiftCode, callingCode, "False\n",
			                            "TestSimpleEnumOptionalCtor", "NewClassCompilerTests", "TomTestSimpleEnumOptionalCtor");
		}
```

stephen-hawley:
Fixed in PR [379](https://github.com/xamarin/binding-tools-for-swift/pull/379)

PR #379, opened 2020-08-13
Trivial enum ctors

stephen-hawley:
Added constructor support for trivial enum ctors, fixing issue [141](https://github.com/xamarin/binding-tools-for-swift/issues/141)

I classify enums into several categories since enums in swift are really discriminated unions. Trivial enums are those that are just homogenous integral cases. These map directly onto C# enums, but because we're talking swift here, there are special cases. You can write methods and constructors on these types. Methods are easy-is - BTfS turns them into extension methods. Constructors? Well, I hadn't considered that initially.

Let's think about what a constructor is from a semantic point of view: it's a static function that takes 0 or more arguments are returns an instance of the type. And given that this is swift, it could also return an optional of that type. Now swift is slightly more complicated than that because constructors also take another argument, which is the type metadata for that type. This argument is *always* ignored unless the type is generic and trivial enums will always be non-generic.

Let's consider the signature of two constructors:
```
public enum TheForce {
    public init (does: Bool) { }
    public init? (does: Int) {}
}
```
The first is really `static TheForce.init (Bool) -> TheForce` and the second is `static TheForce.init(Int) -> TheForce?`.

Seeing this transformation is the key to having these in C#. What I do is the take the `FunctionDeclaration` for the ctor and do the following:
1 - Make it static
2 - Remove the first parameter list (this contains the metadata argument)
3 - Change the name (`.ctor`) to `init`

Side note: early on in dismantling what swift generated, I noted that types had 2 varieties of constructor, one which called the other and destructors. I called them call `.ctor` and `.dtor`. I later learned that these are an allocator and an initializer.

The above transformation happens in `RecastEnumCtorAsStaticFactory`. The result gets put into the same class as any extension methods for the trivial enum and will either get named `Init` or `InitOptional`, but they're just static methods, not actual extensions.

Now comes a traditional engineering tradeoff: where to do this work? There is code already for handling constructors for structs which is very similar. I could just use that code and add in special cases for this or I could copy/paste it, change the issue and risk propagating bugs by poor factoring. Putting in the special case code into the struct handling makes that code more fragile and harder to maintain. As far as I'm concerned, this is a lose-lose proposition. I chose the code duplication. This code loops through all the TopLevelFunction ctors, finds the corresponding `FunctionDeclaration`, recasts it as static, finds wrappers, writes the pinvoke, writes the function declaration and then injects the marshaling code.

In writing all of this I got the main case working, but optional constructors failed. I had intended to do those in another PR, but I was also getting dozens of new failures in existing tests. Why? Because every trivial enum in swift has an implicit constructor that takes in `Int` to build it from the raw value but unless you have up to 64 bits worth of cases, then there are raw values that don't generate correct cases, so the constructor could fail. Any test I had written with a trivial enum was now generating a (broken) optional initializer.

Optional initializers were failing for two reasons:
1 - Typically the return of a ctor is put into a passed in reference since we can't guarantee that the swift and C# calling conventions match - except for heap allocated types and trivial enums. I had missed the trivial enum case in exactly 1 spot in `FindWrapperForConstructor`. Now fixed.
2 - We have never asked the `StructMarshal.Metatypeof` a trivial enum. How do I know this? Because we were getting an exception trying to do that. The type metadata exists and we have a handle to it. We were just missing the code to get it. However, this fix is more subtle than it appears. Most of the code to get a SwiftMetatype object is the same: "is it a foo? If so get it's `SwiftMetatype`. We know definitively that if, say, the object implements `ISwiftObject` there is precisely one way to get it's type metadata. Not so with enums. An enum may be either a swift enum or it may be an ObjC enum. So I adopt the `TryGet` pattern in C# for a swift trivial enum. Turns out that it's pretty much the same as the code for any other value type, just without an exception. I could have rewritten all of that code to be better factored for both but chose to call the value type code from the trivial enum code and swallow the exception. Again, engineering trade-off. If the enum isn't a swift enum, it will fall through (eventually) to the code that picks up the ObjC metadata.

Tests as per usual.

### DIFF
--- a/SwiftReflector/NewClassCompiler.cs
+++ b/SwiftReflector/NewClassCompiler.cs
@@ -1431,6 +1431,7 @@ namespace SwiftReflector {
 
 			ImplementProperties (en, picl, usedPinvokeNames, enumDecl, classContents, null, use, wrapper, false, true, tlf => true, swiftLibraryPath, errors);
 			ImplementMethods (en, picl, usedPinvokeNames, swiftClassName, classContents, enumDecl, use, wrapper, tlf => true, swiftLibraryPath, errors);
+			ImplementTrivialEnumCtors (en, picl, usedPinvokeNames, enumDecl, classContents, use, wrapper, swiftLibraryPath, errors);
 			return en;
 		}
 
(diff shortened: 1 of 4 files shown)


SwiftReflector/NewClassCompiler.cs
```diff
@@ -4477,6 +4478,86 @@ static int MachinePointerSize (MachO.Architectures arch)
...
+					var engine = new MarshalEngine (use, localIdents, TypeMapper, wrapper.Module.SwiftCompilerVersion);
+
+					publicCons.Body.AddRange (engine.MarshalFunctionCall (wrapperFunc, false, pinvokeConsRef,
+						publicCons.Parameters, enumDecl, funcDecl.ReturnTypeSpec, csReturnType, swiftInstanceType: null,
+						instanceType: null, includeCastToReturnType: true, wrapper, includeIntermediateCastToLong: true));
+
+					en.Methods.Add (publicCons);
+					picl.Methods.Add (piConstructor);
+				} catch (Exception err) {
+					var ex = ErrorHelper.CreateError (ReflectorError.kCompilerReferenceBase + 88, err, $"Exception thrown marshaling to swift in enum constructor {recastCtor}");
```
rolfbjarne, line 4555: You're swallowing this exception (nothing happens with the return value from CreateError), is that intentional?

stephen-hawley: 👍

SwiftReflector/NewClassCompiler.cs
```diff
@@ -4477,6 +4478,87 @@ static int MachinePointerSize (MachO.Architectures arch)
...
+					var ex = ErrorHelper.CreateError (ReflectorError.kCompilerReferenceBase + 88, $"Unable to find FunctionDeclaration for wrapper function {wrapperFunction.MangledName} in enum {enumDecl.ToFullyQualifiedName ()}, skipping.");
+					errors.Add (ex);
+					continue;
+				}
+				try {
+					var piConstructor = TLFCompiler.CompileMethod (wrapperFunc, use,
+						libraryPath, wrapperFunction.MangledName, pinvokeConsName, true, true, false);
+
+					var publicCons = TLFCompiler.CompileMethod (recastCtor, use, libraryPath, wrapperFunction.MangledName,
+						consName, isPinvoke: false, isFinal: true, isStatic: true);
```
spouliot, line 4534: ❤️

Issue #119, opened 2018-08-10
[SoM] use of unresolved identifier


chamons:
```
import Darwin
import IOKit.pwr_mgt
import Foundation

public struct System {

    public enum LOAD_AVG {
        case short
        case long
    }
    
    public static func loadAverage(_ type: LOAD_AVG = .long) -> [Double] {
        return []
    }
}
```

```
Error (Fatal): SwiftRepro-System-LOAD_AVG.swift:7:12: error: use of unresolved identifier 'LOAD_AVG'
    return LOAD_AVG.__derived_enum_equals(a, b);
           ^~~~~~~~
Darwin.LOAD_SCALE:1:12: note: did you mean 'LOAD_SCALE'?
public var LOAD_SCALE: Int32 { get }
           ^
Darwin.VM_LOADAVG:1:12: note: did you mean 'VM_LOADAVG'?
public var VM_LOADAVG: Int32 { get } 
```

From SystemKit.framework

stephen-hawley:
Fixed in PR [382](https://github.com/xamarin/binding-tools-for-swift/pull/382)

PR #382, opened 2020-08-17
Fix for nested trivial enums

stephen-hawley:
Made minor adjustment in handling of methods and properties on trivial enums, fixing issue [119](https://github.com/xamarin/binding-tools-for-swift/issues/119).

The problem in the initial issue was that the enum was nested and wrapping code wasn't using the proper path to the type, but that apparently has been solved at some point in the past.

Trivial swift enums are represented by real C# enums and all methods and properties get implemented as extensions. Today I learned that C# doesn't allow extension methods on nested classes, so this doesn't work so well. I put in code to detect this case and still write static methods, just not make them extensions.

How do we know that a `BaseDeclaration` is a nested type? If `this` is a `TypeDeclaration` (parent type of `ClassDeclaration`, `StructDeclaration`, `EnumDeclaration` and `ProtocolDeclaration`) and its `Parent` property is non-null.

Test as per usual.

### DIFF
--- a/tests/tom-swifty-test/SwiftReflector/EnumTests.cs
+++ b/tests/tom-swifty-test/SwiftReflector/EnumTests.cs
@@ -78,5 +78,25 @@ public enum SomeForce {
 			var callingCode = CSCodeBlock.Create (forceDecl, printer);
 			TestRunning.TestAndExecute (swiftCode, callingCode, "False\n");
 		}
+
+		[Test]
+		public void NestedEnum ()
+		{
+			var swiftCode = @"
+public struct System {
+    public enum LOAD_AVG {
+        case short
+        case long
+    }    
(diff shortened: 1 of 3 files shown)


tests/tom-swifty-test/SwiftReflector/EnumTests.cs
```diff
@@ -78,5 +78,25 @@ public enum SomeForce {
...
+		public void NestedEnum ()
+		{
+			var swiftCode = @"
+public struct System {
+    public enum LOAD_AVG {
+        case short
+        case long
+    }    
+    public static func loadAverage(type: LOAD_AVG = .long) -> Int {
+        return 42
```
whitneyschmidt, line 92: 💯

Issue #387, opened 2020-08-19
Need to check T.Type forms in return types and virtual methods


stephen-hawley:
Need to check code of the form:

```
open class Foo<T> {
    public init (x: T) {
        thing = x
    }
    private var thing: T
    open func getThingType () -> T.Type {
         return type (of: thing)
    }
    open func myTypeOf<U>(x: U) -> U.Type {
        return type (of: x)
    }
    open func identityType (t: T.Type) -> T.Type {
        return t
    }
}
```

Issue #109, opened 2018-09-05
[SoM] Unable to find swiftmodule file for T for target


chamons:
```
public extension Dictionary {
    func value<T>(forKey: String, ofType: T.Type) -> T? {
        return nil
    }
}
```
```
Error (Fatal): Unable to find swiftmodule file for T for target x86_64-apple-macosx10.12. (CompilerException)
Error (Fatal): Failed to wrap module SwiftRepro. (CompilerException)
```

stephen-hawley:
Fixed in PR [388](https://github.com/xamarin/binding-tools-for-swift/pull/388)

PR #388, opened 2020-08-19
Better support for type arguments

stephen-hawley:
Added preliminary support for generic Type arguments fixing issue [109](https://github.com/xamarin/binding-tools-for-swift/issues/109)

In this case, we get presented with a generic function on `T` with an argument of type `T.Type`. This was failing in a number of places. The first place was demangling because we had never seen this type before. To support this, I extended the `SwiftClassType` which is a stand-in for type metadata to act as a named metadata type or a generic reference to a metadata type.

Once that was done, I needed to change the predicate for `IsTypeSpecGeneric` because technically `T.Type` is generic. I did this by making a predicate to identify any `TypeSpec` of the form `generic_name.Type` as a generic which also returns the generic name. Similarly, I modded the code to get the generic depth and index to work with this form as well.

Next, I changed the code to match parameter types between `TypeSpec` and `SwiftType` types to understand that a `T.Type` form is not a generic reference. Then I changed the type mapper to understand this as well.

Finally, I had to mod the `SLGenericReferenceType` so that it could generate a `T.Type` form.

I say preliminary because this only addresses one narrow case. This is likely to work correctly in static and non-virtual methods, but is likely to not work in virtual functions and probably not as return values. I opened up an [issue](https://github.com/xamarin/binding-tools-for-swift/issues/387) to look at those.

Tests as per usual.

### DIFF
--- a/SwiftReflector/SwiftXmlReflection/BaseDeclaration.cs
+++ b/SwiftReflector/SwiftXmlReflection/BaseDeclaration.cs
@@ -358,16 +358,42 @@ namespace SwiftReflector.SwiftXmlReflection {
 			return ns != null && IsTypeSpecGeneric (ns.Name);
 		}
 
+		public bool IsTypeSpecGenericMetatypeReference (TypeSpec sp)
+		{
+			if (sp.ContainsGenericParameters)
+				return false;
+			string notUsed;
+			return sp is NamedTypeSpec ns && IsTypeSpecGenericMetatypeReference (ns.Name, out notUsed);
+		}
+
+		public bool IsTypeSpecGenericMetatypeReference (string typeSpecName, out string genericPart)
+		{
+			if (typeSpecName.Contains ('.')) {
+				var parts = typeSpecName.Split ('.');
+				if (parts.Length == 2 && parts [1] == "Type") {
+					genericPart = parts [0];
+					return true;
+				}
+			}
+			genericPart = null;
+			return false;
+		}
+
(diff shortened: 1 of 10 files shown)


SwiftReflector/SwiftXmlReflection/BaseDeclaration.cs
```diff
@@ -358,16 +358,42 @@ public bool IsTypeSpecGenericReference (TypeSpec sp)
 			return ns != null && IsTypeSpecGeneric (ns.Name);
 		}
 
-		public bool IsTypeSpecGeneric (string typeSpecName)
+		public bool IsTypeSpecGenericMetatypeReference (TypeSpec sp)
+		{
+			if (sp.ContainsGenericParameters)
+				return false;
+			string notUsed;
+			return sp is NamedTypeSpec ns && IsTypeSpecGenericMetatypeReference (ns.Name, out notUsed);
```
rolfbjarne, line 366: You can use `var _` for out variables you don't need:

```suggestion
			return sp is NamedTypeSpec ns && IsTypeSpecGenericMetatypeReference (ns.Name, out var _);
```

rolfbjarne (approved):

SwiftReflector/SwiftXmlReflection/BaseDeclaration.cs
stephen-hawley: TIL

Issue #66, opened 2019-05-07
[SoM] Instance type arguments should null check


stephen-hawley:
For any argument that is an instance of a swift class, the argument should be null checked before getting passed in.

premarshal code for an argument should probably be something like:
if (p == null)
    throw new ArgumentNullException (nameof(p));

PR #413, opened 2020-08-27
Added null checks to arguments

stephen-hawley:
Added code to generate null argument checks fixing issue [66](https://github.com/xamarin/binding-tools-for-swift/issues/66)

First I added `nameof` to `CSFunctionCall` and while I was in there, I refactored all the `_of` functions to all go through the same code, getting rid of duplicate code.

Second I exposed `ThrowOnNull` as public. There is another copy of this in the SwiftReflector project and since SwiftReflector has a reference to SwiftRuntimeLibrary, I should get rid of it and refactor. I didn't do this because that's potentially a big change and belongs in a PR of its own (ooh! premonition!).

The actual change happens in `MarshalEngine`. There's a bottleneck for marshaling `ISwiftObject` derived types and one for `ISwiftNominalType` derived types. This makes it really easy to make this a uniform change. 

TIL that `nameof(this)` in C# is meaningless, which is weird because it does have a name. No matter - not generating a null test for `this` is the right thing anyway since `this` damn well better be non-null. That particular code generated 392 test failures, so I think the coverage is pretty good here.

Finally for the test specific to this, I test a positive failure by trying to pass null to a function that takes an `ISwiftObject` to see what happens.

### DIFF
--- a/SwiftRuntimeLibrary/Exceptions.cs
+++ b/SwiftRuntimeLibrary/Exceptions.cs
@@ -3,7 +3,7 @@
 
 using System;
 namespace SwiftRuntimeLibrary {
-	public static class Exceptions {
+	internal static class Exceptions {
 		public static T ThrowOnNull<T> (T o, string name, string message = null) where T : class
 		{
 			name = name ?? "::no name supplied::";
--- a/SwiftRuntimeLibrary/Exceptions.cs
+++ b/SwiftRuntimeLibrary/Exceptions.cs
@@ -3,7 +3,7 @@
 
 using System;
 namespace SwiftRuntimeLibrary {
-	internal static class Exceptions {
+	public static class Exceptions {
 		public static T ThrowOnNull<T> (T o, string name, string message = null) where T : class
 		{
 			name = name ?? "::no name supplied::";
--- a/tests/tom-swifty-test/SwiftReflector/NewClassCompilerTests.cs
+++ b/tests/tom-swifty-test/SwiftReflector/NewClassCompilerTests.cs
@@ -1468,6 +1468,36 @@ public struct PitchSet
 			var callingCode = CSCodeBlock.Create (printer);
 			TestRunning.TestAndExecute (swiftCode, callingCode, "got here\n");
 		}
+
+		[Test]
+		public void DontPassMeNull ()
+		{
+			var swiftCode = @"
+public class EasyToRepresent
+{
+	public init () {}
+}
+
+public func reportIt (a: EasyToRepresent) -> Bool {
+	return true;
+}
+";
+
+			// try {
+			//    TopLevelEntities.ReportIt (null);
+			// } catch (ArgumentNullException err) {
+			//    Console.WriteLine("Here." + err.Message);
+			// }
+
+			var errID = new CSIdentifier ("err");
+			var callIt = CSFunctionCall.FunctionCallLine ("TopLevelEntities.ReportIt", false, CSConstant.Null);
+			var printer = CSFunctionCall.ConsoleWriteLine (CSConstant.Val ("Here.") + errID.Dot(new CSIdentifier ("Message")));
+			var tryCatch = new CSTryCatch (CSCodeBlock.Create (callIt),
+				typeof (ArgumentNullException), errID.Name, CSCodeBlock.Create (printer));
+			var callingCode = CSCodeBlock.Create (tryCatch);
+
+			TestRunning.TestAndExecute (swiftCode, callingCode, "Here.Value cannot be null.\nParameter name: a\n");
+		}
 	}
 }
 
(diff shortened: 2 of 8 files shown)


tests/tom-swifty-test/SwiftReflector/NewClassCompilerTests.cs
```diff
@@ -1468,6 +1468,36 @@ public struct PitchSet
...
+public func reportIt (a: EasyToRepresent) -> Bool {
+	return true;
+}
+";
+
+			// try {
+			//    TopLevelEntities.ReportIt (null);
+			// } catch (ArgumentNullException err) {
+			//    Console.WriteLine("Here." + err.Message);
+			// }
```
spouliot, line 1490: In interop/binding code a single null check might not be enough. The managed representation of `EasyToRepresent` can be disposed, which can mean it's handle will be reset.

You can test this with
```csharp
var e2r = new EasyToRepresent ();
e2r.Dispose ();
TopLevelEntities.ReportIt (e2r);
```

If you get a `ObjectDisposedException` then this is already handled elsewhere. If you get a native crash because of a null value then your checks are incomplete.

SwiftRuntimeLibrary/Exceptions.cs
```diff
@@ -3,7 +3,7 @@
 
 using System;
 namespace SwiftRuntimeLibrary {
-	internal static class Exceptions {
+	public static class Exceptions {
```
spouliot, line 6: This helper method cannot be inlined since it throws. It means all null checks (cheap) now requires a call (less cheap).

That might not be obvious to callers, which makes it a _not so good_ candidate to be a public API where the impact cannot be known.

OTOH it could be fixed by moving the actual throws to another method... which would make inlining possible.

stephen-hawley: If I do that then it obviates the benefit for using this.

tests/tom-swifty-test/SwiftReflector/NewClassCompilerTests.cs
stephen-hawley: Good point - but that check will happen when the marshaler tries to do a `Retain` on the object, so it's already there.

SwiftRuntimeLibrary/Exceptions.cs
spouliot: Which benefit ?

The source/call remains identical `Exceptions.ThrowOnNull (foo, "foo");` on the caller side.

It's the `ThrowOnNull` implementation that changes, being split in two. The first part only check for null and call a `Throw*` method if needed. Since the method won't be throwing (by itself) it can be inlined (at the JIT's discretion).

If an exception occurs then a call will be needed - but that's already in a non-performant path so an additional call is unlikely to be noticeable.

tests/tom-swifty-test/SwiftReflector/NewClassCompilerTests.cs
stephen-hawley: Actually, I just check `Retain` and it needs an `IntPtr.Zero` check - opened an [issue](https://github.com/xamarin/binding-tools-for-swift/issues/414).

spouliot: ok, then the test can wait the fix - since it would not be helpful right now

SwiftRuntimeLibrary/Exceptions.cs
stephen-hawley: The benefit is to make it more compact. There are some internal places where I use it specifically for the fluent implementation (eg, `base(Exceptions.ThrowOnNull(arg, nameof (arg)),...)` but if this takes away inlining in the common case, there's no point.

spouliot:
```
base(Exceptions.ThrowOnNull(arg, nameof (arg)),...)
```

hmm... that would still work with what I said.

IOW the `ThrowOnNull` API (at least its signature) does not have to change.

But it's implementation does need to change. If it does not throw then it can be inlined, but it does not mean the throwing goes back to the caller. There just need to be another (private) method that does the throw.

```csharp
// this method does not throw and the JIT can (at its discretion) inline it
// that can be important since a `call` can be much costlier than a `null` check
public static T ThrowOnNull<T> (T o, string name, string message = null) where T : class
 {
	if (o == null)
		ThrowNullException (name ?? "::no name supplied::", message);
	return o;
}

// this method does throw and the JIT cannot inline it
// however it's not inside the normal, where performance might matter, path
static void ThrowNullException (string name, string message)
{
	if (message == null)
		throw new ArgumentNullException (name);
	else
		throw new ArgumentNullException (name, message);
}
```

stephen-hawley: OK - I see that. I can always add that to the implementation later.

Issue #414, opened 2020-08-27
Retain should check for IntPtr.Zero


stephen-hawley:

stephen-hawley:
Addressed in PR [419](https://github.com/xamarin/binding-tools-for-swift/pull/419)

PR #419, opened 2020-08-31
notify on use of disposed object

stephen-hawley:
Check for 0 handle in retain fixing issue [414](https://github.com/xamarin/binding-tools-for-swift/issues/414)

Step 1:
Ensure that the handle gets set to `IntPtr.Zero` on Dispose
Step 2:
🤷‍♀️ 
Step 3: ~~Profit~~ check for `IntPtr.Zero` in `RetainSwiftObject`

Why does this work?
`IntPtr.Zero` is never a valid value for a swift class instance. If the object gets Disposed, setting the handle to `IntPtr.Zero` is the ideal value for "no good"
Why don't you check to see if `Dispose` was called?
Because `SwiftRetainObject` gets called before any handle gets passed to swift and it doesn't have access to the flag. In order to do that, I'd have to do one of the following:
1 - change the code generator to inject checks before calls to `RetainSwiftObject`, which also may not have access to the object's state
2 - put a disposed flag into a deliciously small interface which has a lot of fan-out of changes

Nope, this is small and relatively safe. I like those traits.

Test passes as expected.

### DIFF
--- a/SwiftRuntimeLibrary/SwiftMarshal/StructMarshal.cs
+++ b/SwiftRuntimeLibrary/SwiftMarshal/StructMarshal.cs
@@ -26,6 +26,8 @@ namespace SwiftRuntimeLibrary.SwiftMarshal {
 			if (obj == null)
 				return IntPtr.Zero;
 			byte* valueBuffer = stackalloc byte [3 * IntPtr.Size];
+			if (obj.SwiftObject == IntPtr.Zero)
+				throw new SwiftRuntimeException ("SwiftObject handle is IntPtr.Zero, likely because it was disposed.");
 			SwiftCore.swift_beginAccess (obj.SwiftObject, valueBuffer, (uint)SwiftExclusivityFlags.Track, IntPtr.Zero);
 			var result = SwiftCore.Retain (obj.SwiftObject);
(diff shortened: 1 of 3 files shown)


SwiftRuntimeLibrary/SwiftMarshal/StructMarshal.cs
```diff
@@ -26,6 +26,8 @@ public static unsafe IntPtr RetainSwiftObject (ISwiftObject obj)
 			if (obj == null)
 				return IntPtr.Zero;
 			byte* valueBuffer = stackalloc byte [3 * IntPtr.Size];
+			if (obj.SwiftObject == IntPtr.Zero)
+				throw new SwiftRuntimeException ("SwiftObject handle is IntPtr.Zero, likely because it was disposed.");
```
spouliot, line 30: minor: I'd put checks/exceptions before any work/allocations (not that it would leak in this case, nor that performance would suffer)

stephen-hawley: 👍

rolfbjarne (approved):

Issue #48, opened 2019-06-18
[SoM] Not handling associated type descriptors


stephen-hawley:
```
_$s12RowValueType6Eureka04RuleC0PTl
_$s23PresentedControllerType6Eureka012PresenterRowC0PTl
_$s9InlineRow6Eureka0aB4TypePTl
```

stephen-hawley:
Fixed in this [PR](https://github.com/xamarin/binding-tools-for-swift/pull/425)

PR #425, opened 2020-09-01
Added associated type descriptors

stephen-hawley:
Added support for demangling associated type descriptor fixing issue [48](https://github.com/xamarin/binding-tools-for-swift/issues/48)

An associated type descriptor looks like this:
```
->AssociatedTypeDescriptor
  ->DependentAssociatedTypeRef ("AssociatedTypeName")
    ->Type
      ->Protocol
        ->Module ("ModuleName")
        ->Identifier ("ProtocolName")
```

Everything from the `Type` node down we can convert directly, so all we need to do is pull out the associated type name.
Defined a new class in the `TL` hierarchy to contain the data and done.

Added 3 tests, which pass.

### DIFF
--- a/SwiftReflector/Demangling/TLDefinition.cs
+++ b/SwiftReflector/Demangling/TLDefinition.cs
@@ -232,5 +232,15 @@ namespace SwiftReflector.Demangling {
 		}
 		public SwiftClassType ProtocolRequirement { get; private set; }
 	}
+
+	public class TLAssociatedTypeDescriptor : TLClassElem {
+		public TLAssociatedTypeDescriptor (string mangledName, SwiftName module, SwiftClassType protocol, SwiftName associatedTypeName, ulong offset)
+			: base (CoreCompoundType.AssociatedTypeDescriptor, mangledName, module, protocol, offset)
+		{
+			AssociatedTypeName = associatedTypeName;
+		}
+
+		public SwiftName AssociatedTypeName { get; private set; }
+	}
 }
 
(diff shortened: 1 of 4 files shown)


SwiftReflector/Demangling/TLDefinition.cs
```diff
@@ -232,5 +232,15 @@ public TLBaseConformanceDescriptor (string mangledName, SwiftName module, SwiftC
...
+
+	public class TLAssociatedTypeDescriptor : TLClassElem {
+		public TLAssociatedTypeDescriptor (string mangledName, SwiftName module, SwiftClassType protocol, SwiftName associatedTypeName, ulong offset)
+			: base (CoreCompoundType.AssociatedTypeDescriptor, mangledName, module, protocol, offset)
+		{
+			AssociatedTypeName = associatedTypeName;
+		}
+
+		public SwiftName AssociatedTypeName { get; private set; }
+	}
```
mandel-macaque, line 244: Little trick to type less:

```csharp
public class TLAssociatedTypeDescriptor : TLClassElem {
		public TLAssociatedTypeDescriptor (string mangledName, SwiftName module, SwiftClassType protocol, SwiftName associatedTypeName, ulong offset)
			: base (CoreCompoundType.AssociatedTypeDescriptor, mangledName, module, protocol, offset)
				=> AssociatedTypeName = associatedTypeName;

		public SwiftName AssociatedTypeName { get; }
	}
```
Don't ask me when it was added in the language, but is nice for getter only properties.

stephen-hawley: 👍

rolfbjarne (approved):

Issue #77, opened 2019-02-11
[SoM] Add a base class for ISwiftObject and ISwiftNominalType


rolfbjarne:
We should implement a base class for ISwiftObject and ISwiftNominalType to reduce code duplication.

See discussion:

https://xamarinhq.slack.com/archives/C9WBDKG93/p1548066602158800
https://xamarinhq.slack.com/archives/C9WBDKG93/p1548166293173700

stephen-hawley:
Inserting the conversation:
I think we should add a common base class for swift-related objects so that we can avoid code duplication (and avoid errors when code isn't duplicated correctly)
I've found this multiple times: a common base class with shared code would avoid such errors we can easily make it so that the base class isn't required for people who don't want to use it for whatever reason
```
public abstract class SwiftNativeValue : IDisposable {
    ~SwiftNativeValue ()
    {
        Dispose (false);
    }
    void Dispose ()
    {
        Dispose (true);
        GC.SuppressFinalize (this);
    }
    protected abstract void Dispose (bool disposing);
}
```
we could also add something for nominal types:
```
public abstract class SwiftNativeNominalObject : SwiftNativeValue, ISwiftNominalType {
    protected void Dispose (bool disposing)
    {
         StructMarshal.Marshaler.ReleaseNominalData (this);
    }
    public byte[] SwiftData { get; set; }
}
```
this would remove a lot of the generated boiler plate code
swift objects would benefit too:
```
public class abstract SwiftNativeObject : SwiftNativeValue, ISwiftObject {
    IntPtr handle;
    SwiftMetatype class_handle;    
    SwiftObjectFlags object_flags = SwiftObjectFlags.IsSwift;

    protected SwiftNativeObject(IntPtr handle, SwiftMetatype classHandle, SwiftObjectRegistry registry)
    {
        if (SwiftNativeObjectAttribute.IsSwiftNativeObject(this))
        {
            object_flags |= SwiftObjectFlags.IsDirectBinding;
        }
        class_handle = classHandle;
        SwiftObject = handle;
        registry.Add(this);
    }

    protected virtual void Dispose(bool disposing)
    {
        if ((object_flags & SwiftObjectFlags.Disposed) != SwiftObjectFlags.Disposed)
        {
            if (disposing)
            {
                DisposeManagedResources();
            }
            DisposeUnmanagedResources();
            SwiftObjectRegistry.Registry.RemoveAndWeakRelease(this);
[...]

PR #432, opened 2020-09-04
moving common code for ISwiftObject

stephen-hawley:
As per issue [77](https://github.com/xamarin/binding-tools-for-swift/issues/77), moving common code into base classes.

This PR includes the implementation for heap allocated objects but doesn't change the generated code (yet).

Notes:
I'm using `SwiftNativeObject` for any _class based_ swift type and `SwiftNativeInstance` for its parent class, representing any swift type that can be instantiated. This differs from the suggestion in the PR, because I want to move away from the term `NominalType` for enums and structs since it is incorrect. I'm going to be shifting to `ValueType` instead, so the struct/enum version will be `SwiftNativeValueType` instead of `SwiftNativeNominalObject`.

Q: should I implement the finalizer in children of `SwiftNativeObject`? I currently am - I'm not sure if its redundant.

### DIFF
--- a/SwiftRuntimeLibrary/SwiftNativeObject.cs
+++ b/SwiftRuntimeLibrary/SwiftNativeObject.cs
@@ -1,0 +1,53 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using SwiftRuntimeLibrary.SwiftMarshal;
+
+namespace SwiftRuntimeLibrary {
+	public abstract class SwiftNativeObject : SwiftNativeInstance, ISwiftObject {
+	IntPtr handle;
+	SwiftMetatype class_handle;
+	SwiftObjectFlags object_flags = SwiftObjectFlags.IsSwift;
+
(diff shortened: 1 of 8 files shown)


chamons (approved):

mandel-macaque (approved):

SwiftRuntimeLibrary/SwiftNativeObject.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using SwiftRuntimeLibrary.SwiftMarshal;
+
+namespace SwiftRuntimeLibrary {
+	public abstract class SwiftNativeObject : SwiftNativeInstance, ISwiftObject {
+	IntPtr handle;
```
rolfbjarne, line 9: The indentation is wrong here.

rolfbjarne (approved):
Looks great!

SwiftRuntimeLibrary/SwiftNativeObject.cs
stephen-hawley: 👍

PR #436, opened 2020-09-11
No longer implementing all of ISwiftObject

stephen-hawley:
Two changes: first did a rename of `SwiftNativeObjectAttribute` to `SwiftNativeObjectTagAttribute`. The reason being that there exists a class `SwiftNativeObject` and code that gets generated often looks like this:
```
[SwiftNativeObject]
public class Foo : SwiftNativeObject {
}
```
and that's confusing. Lesson: naming is hard.

The rest was removing the code that generates `ISwiftObject` requirements. Previously, I had a "standard" constructor with the signature `SomeClass (IntPtr handle, SwiftObjectRegistry registry)`, but it was better to have the signature `protected SomeClass (IntPtr handle, SwiftMetatype classHandle, SwiftObjectRegistry registry)`.
"Hey Steve, since the SwiftObjectRegistry is a singleton, why don't you lost that argument?" Because it's quite possible for a swift constructor to have the signature `SomeClass (IntPtr handle, SwiftMetatype classHandle)` and we don't want to have to deal with that conflict, whereas `SwiftObjectRegistry` should never appear in a Swift signature.

### DIFF
--- a/SwiftReflector/MarshalEngine.cs
+++ b/SwiftReflector/MarshalEngine.cs
@@ -113,7 +113,8 @@ namespace SwiftReflector {
 				} else {
 					if (isHomonym) {
 						use.AddIfNotPresent (typeof (SwiftObjectRegistry));
-						postMarshalCode.Add (CSReturn.ReturnLine (new CSFunctionCall (cl.Name.Name, true, thisIntPtr, new CSIdentifier ("SwiftObjectRegistry.Registry"))));
+						postMarshalCode.Add (CSReturn.ReturnLine (new CSFunctionCall (cl.Name.Name, true, thisIntPtr,
+							new CSFunctionCall ($"{cl.Name.Name}.GetSwiftMetatype", false), new CSIdentifier ("SwiftObjectRegistry.Registry"))));
 					} else {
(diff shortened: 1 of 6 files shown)


spouliot (approved):
At some point you should add a "Generator Diff" like we have on xamarin-macios that compare the generated code changes. That's very useful to see the impact of the PR, both for the author and reviewers.

c.c. @rolfbjarne

chamons (approved):

stephen-hawley:
That's a great idea - I'll do that for the next PR which will be similar in scope.

rolfbjarne (approved):

rolfbjarne:
> "Hey Steve, since the SwiftObjectRegistry is a singleton, why don't you lost that argument?" Because it's quite possible for a swift constructor to have the signature `SomeClass (IntPtr handle, SwiftMetatype classHandle)` and we don't want to have to deal with that conflict, whereas `SwiftObjectRegistry` should never appear in a Swift signature.

Today I learned something new! But could you add ^ to the code as a comment somewhere? This is both important and non-obvious at the same time, which means that it should be a bit more visible than in a commit message it's unlikely anybody will read ever again.

stephen-hawley:
Comments added to `SwiftNativeObject`, @rolfbjarne

Issue #155, opened 2018-04-10
Swift-o-matic properties need to support homonyms


stephen-hawley:
swift allows properties that are have the same name but different types:
```
public class Foo {
    public static var X:Int = 3
    public var X:Float = 4
}
```
swift-o-matic needs to support that.

stephen-hawley:
Actually, probably right issue. Double check.

stephen-hawley:
Fixed in the [PR](https://github.com/xamarin/binding-tools-for-swift/pull/450)

PR #450, opened 2020-09-22
Handle difference between static and instance properties with the same name

stephen-hawley:
Addresses issue [155](https://github.com/xamarin/binding-tools-for-swift/issues/155)

I want to say up front that the problems here were totally **not** where I expected them to be.

Here's the setting. This is legal in swift:
```
public class Foo {
    public init () { }
    public static var X: Int = 3
    public var X: Float = 4.0
}
```

You can't have properties that are the same staticicity with the same name, but if one is static and the other is not, why yes, that's just fine thank you.

I fully expected this to break in the C# code. Reader, it did not.

There were two issues in that the wrapper name for static and non-static properties conflicted and the second was that the methods `GetGetter` and `GetSetter` weren't matching on staticicity.

Step 1 was to refactor all the flavors of `WrapperName` to understand when a function or property is static and it changes the caps of various things. Note that I don't change the marker for extensions because they're always effectively static anyway.

Step 2 was to change `SearchForPropertyAccessor` to do the right thing with regards to statics.

And much to my delight, the C# code works out of the box because I apparently already had code to disambiguate duplicate property names. Hooray!

### DIFF
--- a/SwiftReflector/MethodWrapping.cs
+++ b/SwiftReflector/MethodWrapping.cs
@@ -85,15 +85,16 @@ namespace SwiftReflector {
 			return $"{kXamPrefix}{classPrefix}{operatorType}{operatorName}";
 		}
 
-		public static string WrapperName (SwiftClassName name, string methodName, bool isExtension)
+		public static string WrapperName (SwiftClassName name, string methodName, bool isExtension, bool isStatic)
 		{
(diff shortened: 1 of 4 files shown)


rolfbjarne (approved):

spouliot (approved):

chamons (approved):
Happy surprise for once!

PR #455, opened 2020-10-06
Update .gitignore.

rolfbjarne:


stephen-hawley:
Are you sure you want to do this? `apple` is a symbolic link to the entire swift source tree.

Issue #462, opened 2020-10-13
Static functions in library-evolution not demangling


stephen-hawley:
Example mangled name: `_$s21NewClassCompilerTests06Publicb4OpenB15MethodBoolFalseC5thingSbyFZTj`

This turns into this tree:
```
kind=Global
  kind=DispatchThunk
    kind=Static
      kind=Function
        kind=Class
          kind=Module, text="NewClassCompilerTests"
          kind=Identifier, text="PublicClassOpenClassMethodBoolFalse"
        kind=Identifier, text="thing"
        kind=Type
          kind=FunctionType
            kind=ArgumentTuple
              kind=Type
                kind=Tuple
            kind=ReturnType
              kind=Type
                kind=Structure
                  kind=Module, text="Swift"
                  kind=Identifier, text="Bool"
```

PR #463, opened 2020-10-13
static function thunks

stephen-hawley:
This addresses static function dispatch thunks in the same way as for properties, fixing issue [462](https://github.com/xamarin/binding-tools-for-swift/issues/462)

I renamed `ConvertToThunk` -> `ConvertToDispatchThunk` to be more clear.
I gave `ConvertDispatchThunk` its own method since it's too complicated to be in a switch
Dispatch thunks for variables are intentionally left out until I encounter them.

This fixes a third of the failing tests in `NewClassCompilerTests.cs`

### DIFF
--- a/SwiftReflector/Demangling/Swift5NodeToTLDefinition.cs
+++ b/SwiftReflector/Demangling/Swift5NodeToTLDefinition.cs
@@ -103,7 +103,7 @@ namespace SwiftReflector.Demangling {
 				new MatchRule {
 					Name = "DispatchThunk",
 					NodeKind = NodeKind.DispatchThunk,
-					Reducer = ConvertToThunk,
+					Reducer = ConvertToDispatchThunk,
 					ChildRules = new List<MatchRule> () { }
 				},
 				new MatchRule {
@@ -778,8 +778,9 @@ namespace SwiftReflector.Demangling {
 			case NodeKind.MaterializeForSet:
 			case NodeKind.WillSet:
 			case NodeKind.ModifyAccessor:
-			case NodeKind.DispatchThunk:
 				return ConvertFunctionProp (node);
+			case NodeKind.DispatchThunk:
+				return ConvertDispatchThunk (node);
 			case NodeKind.Variable:
 				return ConvertVariable (node, false);
 			case NodeKind.PropertyDescriptor:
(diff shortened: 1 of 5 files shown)


chamons (approved):

SwiftReflector/Demangling/Swift5NodeToTLDefinition.cs
```diff
@@ -833,6 +834,40 @@ public TLDefinition Convert (Node node)
 			}
 		}
 
+
+		TLDefinition ConvertDispatchThunk (Node node)
+		{
+			switch (node.Children [0].Kind) {
+			case NodeKind.Getter:
```
tj-devel709, line 841: Question: why does this switch not need 'breaks'?

stephen-hawley: Every includes a return.

tj-devel709 (approved):

SwiftReflector/Demangling/Swift5NodeToTLDefinition.cs
mandel-macaque: @tj-devel709 you will get a compiler error if is not correct: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0163

mandel-macaque (approved):
Approving, small comment to reduce code branching.

SwiftReflector/Inventory/FunctionInventory.cs
```diff
@@ -30,6 +30,46 @@ public override void Add (TLDefinition tld, Stream srcStm)
 			overloads.Add (tlf, srcStm);
 		}
 
+		public TLFunction ContainsEquivalentFunction (TLFunction func)
+		{
+			// for finding a func with the same name and signature which may or may not be a thunk
+			var nameMatched = MethodsWithName (func.Signature.Name);
+			if (nameMatched.Count == 0)
+				return null;
```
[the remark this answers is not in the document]
stephen-hawley: Tough call; integer comparison is super cheap and avoids the allocation of an enumerator if there are no matches.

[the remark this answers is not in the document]
mandel-macaque: Depends if the enumerator needs to do the count :)

Issue #468, opened 2020-10-14
Functions not handled with dispatch thunk


stephen-hawley:
Typical example `_$s21NewClassCompilerTests4FoodP9AddOnionsyyFTj` which generates the following tree:
```
kind=Global
  kind=DispatchThunk
    kind=Function
      kind=Protocol
        kind=Module, text="NewClassCompilerTests"
        kind=Identifier, text="Food"
      kind=Identifier, text="AddOnions"
      kind=Type
        kind=FunctionType
          kind=ArgumentTuple
            kind=Type
              kind=Tuple
          kind=ReturnType
            kind=Type
              kind=Tuple
```

PR #469, opened 2020-10-15
Thunk all the things!

stephen-hawley:
This PR fixes issue [468](https://github.com/xamarin/binding-tools-for-swift/issues/468) where most function dispatch thunks were being ignored.

This PR involves a big refactoring of the `SwiftBaseFunctionType` sub hierarchy to clean everything up. Well, at least as much as typical OO code will allow.

Every function type can (potentially) have a thunk associated with it, so I hoisted this notion into the thunk base class by adding an `IsThunk` property as well as an abstract `AsThunk` method which returns the thunk equivalent of the type. This would actually be a great place to use dynamic self if we had that in C#. I also briefly considered making the call be generic `T AsThunk<T>()` but decided that that doesn't really give us much in context.

The end result is that the handling of dispatch thunks in the demangler is much cleaner as is thunks in function handling.


rolfbjarne (approved):

dalexsoto (approved):

chamons (approved):

jingliancui:
awesome

PR #472, opened 2020-10-16
[devicetests] Enabling Provisioning Profiles

tj-devel709:
This change allows azure-pipelines.yml to access Maccore provisioning profiles, give us results from the tests, and changes one of the tests to mac only.
-With help from Manuel

### DIFF
--- a/jenkins/azure-pipelines.yml
+++ b/jenkins/azure-pipelines.yml
@@ -125,13 +125,14 @@ jobs:
     condition: succeededOrFailed()
   
   - bash:
+      pwd
       ls -la
     displayName: 'Check Path for artifact'
 
   - task: PublishPipelineArtifact@1
     displayName: 'Publish Artifact: TestResult'
     inputs:
-      targetPath: 'tests/TestResult.md'
+      targetPath: 'binding-tools-for-swift/tests/TestResult.md'
       artifactName: TestResults
     continueOnError: true
     condition: succeededOrFailed()
--- a/jenkins/azure-pipelines.yml
+++ b/jenkins/azure-pipelines.yml
@@ -126,7 +126,7 @@ jobs:
   - task: PublishPipelineArtifact@1
     displayName: 'Publish Artifact: TestResult'
     inputs:
-      targetPath: $(System.DefaultWorkingDirectory)/binding-tools-for-swift/tests
+      targetPath: $(System.DefaultWorkingDirectory)/binding-tools-for-swift/tests/TestResults.md
       artifactName: TestResults
     continueOnError: true
     condition: succeededOrFailed()
--- a/jenkins/azure-pipelines.yml
+++ b/jenkins/azure-pipelines.yml
@@ -30,6 +30,7 @@ jobs:
   steps:
   - checkout: maccore
     persistCredentials: true
+  - checkout: self
 
   - bash: |
       set -x
--- a/jenkins/azure-pipelines.yml
+++ b/jenkins/azure-pipelines.yml
@@ -98,9 +98,6 @@ jobs:
       set -e
       set -x
       ./binding-tools-for-swift/jenkins/build-package.sh
-      ls -la
-      ls .. -la
-      ls ../.. -la
       ls -la ./package || true
     displayName: 'Package'
 
--- a/jenkins/azure-pipelines.yml
+++ b/jenkins/azure-pipelines.yml
@@ -125,8 +125,8 @@ jobs:
   - task: PublishPipelineArtifact@1
     displayName: 'Publish Artifact: TestResult'
     inputs:
-      targetPath: $(System.DefaultWorkingDirectory)/binding-tools-for-swift/tests/tom-swifty-test/TestResult.xml
-      artifactName: TestResults
+      targetPath: $(System.DefaultWorkingDirectory)/binding-tools-for-swift/tests/tom-swifty-test/TestResult.md
+      artifactName: TestResult
     continueOnError: true
     condition: succeededOrFailed()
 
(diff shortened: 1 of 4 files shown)


tj-devel709:
Please ignore how many commits there are.. Also my vm thinks it is October 9th even though I've changed the date multiple times hence the commits from today saying they are 8 days old.

jenkins/azure-pipelines.yml
```diff
@@ -76,16 +104,33 @@ jobs:
...
+    continueOnError: true
+    condition: succeededOrFailed()
+
+  - task: PublishPipelineArtifact@1
+    displayName: 'Publish Artifact: TestResult'
+    inputs:
+      targetPath: $(System.DefaultWorkingDirectory)/binding-tools-for-swift/tests/TestResults.md
+      artifactName: TestResults
+    continueOnError: true
+    condition: succeededOrFailed()
```
whitneyschmidt, line 132: This looks great! The only thing I'd be extra careful with is checking that all of your conditions run as expected. From the docs, `Only when all previous dependencies have succeeded. This is the default if there is not a condition set in the YAML.` For example, if "upload packages" fails would you still want to run the tests? (as-is, I *think* that you'd skip to "publish unit device test results" since that runs regardless of previous success/failure.

I ran into some issues when I was in YAML b/c I didn't explicitly specify the condition for some steps so tasks stopped running earlier in the process than we wanted :(

https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml

whitneyschmidt (approved):
After double-checking conditions.

jenkins/azure-pipelines.yml
tj-devel709: Good catch! I'm not too sure at the moment which dependencies the tests need. I am assuming uploading the packages are not too important to the tests but I am guessing the steps previous to that are important for tests. I will look further into that. Thanks!

tests/tom-swifty-test/SwiftRuntimeLibraryTests/SwiftDateTests.cs
```diff
@@ -17,6 +17,8 @@ namespace SwiftRuntimeLibraryTests {
 	[TestFixture]
 	public class SwiftDateTests {
 
+#if _MAC_TS_TEST_
```
rolfbjarne, line 20: When is this set?

tj-devel709: I'm sorry I don't think I understand your question. I added this because to my understanding, this is a macOS specific test that was trying to be run on iOS device (VM) so it was causing errors. I used the same '#if _MAC_TS_TEST_' that some tests in the 'binding-tools-for-swift/tests/tom-swifty-test/SwiftReflector' use to specify a test for mac. Is there a different way I should do this?

stephen-hawley: IIRC, `MAC_TS_TEST` only gets used inside code that was **generated** by a test.

tj-devel709: Actually let me double check

stephen-hawley: Isn't it already being pinned to `macOS` with `platform: platform`?

tj-devel709: So without the `#if MAC_TS_TEST_`, binding-tools-for-swift/tests/tom-swifty-test/SwiftRuntimeLibraryTests/SwiftDateTests.cs was segfaulting as seen here: https://dev.azure.com/xamarin/public/_build/results?buildId=28617&view=logs&j=b944efd9-1a3a-54bf-926f-a01c570708bc&t=a206e760-4e0d-53e1-fe50-36a17b4c1b3f&l=22164 so it seemed to me that it was trying to run this test on iOS? @stephen-hawley

stephen-hawley: Right, but it's never being run on the device. That looks like it's failing a leak test. What you did was comment the test out entirely. I think that you probably want to ignore this test and open an issue to look at it later.

tj-devel709: Okay sounds good!

tj-devel709: Removed the changes to this test!

rolfbjarne (approved):

mandel-macaque (changes requested):
Do not check in the test result file, AFAIK is not needed and will be generated

tj-devel709:
> Do not check in the test result file, AFAIK is not needed and will be generated

Are you saying the file is not needed or just don't put anything inside the file?

stephen-hawley (approved):

jenkins/azure-pipelines.yml
```diff
@@ -8,13 +8,41 @@ trigger:
...
+resources:
+  repositories:
+  - repository: maccore
+    type: github
+    name: xamarin/maccore
+    ref: refs/heads/main
+    endpoint: xamarin
+
+variables:
+  -group: xamops-azdev-secrets
```
mandel-macaque, line 20: Missing space between - and group. should be '- group:'

mandel-macaque (changes requested):

jenkins/azure-pipelines.yml
tj-devel709: Good catch!

mandel-macaque (approved):

Issue #493, opened 2020-10-29
Integrate bug fix from Apple for top-level let bindings


stephen-hawley:
See [this issue here](https://bugs.swift.org/browse/SR-13790).

And remove `Ignore` for the following tests:
`TheEpsilonIssue`
`PropertyWrapTests.ClosureTLProp`

PR #494, opened 2020-10-29
bug in swift compiler rears it's less than appealing head

stephen-hawley:
This PR doesn't fix an issue but hides it.

There were two tests that were failing from swift code that looks like this:
```
public let e = 2.718
```

When we run the reflector on this, we were being told that the `e` was a `var` and not a `let` and that it had both a valid getter and a valid setter.

Ultimately the cause for this is...swift!
When we compile the module, it writes a `.swiftinterface` file that looks like this:
```
import Swift
public var e: Swift.Double
```
which is clearly incorrect. Handling this is completely out of my control and I've filed a bug with Apple [here](https://bugs.swift.org/browse/SR-13790). There is no workaround for this. I added a couple new unit tests, but these don't generate `.swiftinterface` files, so they pass as is. I've temporarily ignored the failing unit tests and opened [an issue](https://github.com/xamarin/binding-tools-for-swift/issues/493) to take care of these when Apple fixes this problem and we have integrated it.

### DIFF
--- a/tests/tom-swifty-test/SwiftReflector/PropertyWrapTests.cs
+++ b/tests/tom-swifty-test/SwiftReflector/PropertyWrapTests.cs
@@ -190,6 +190,7 @@ public var DoubleStr: String {
 		}
 
 		[Test]
+		[Ignore ("apple's top level let issue")]
 		public void ClosureTLProp ()
 		{
 			var swiftCode = @"
--- a/tests/tom-swifty-test/SwiftReflector/PropertyWrapTests.cs
+++ b/tests/tom-swifty-test/SwiftReflector/PropertyWrapTests.cs
@@ -190,7 +190,7 @@ public var DoubleStr: String {
 		}
 
 		[Test]
-		[Ignore ("apple's top level let issue")]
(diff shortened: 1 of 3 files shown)


tests/tom-swifty-test/SwiftReflector/PropertyWrapTests.cs
```diff
@@ -190,6 +190,7 @@ public void ComputedPropertyTestString ()
 		}
 
 		[Test]
+		[Ignore ("apple's top level let issue")]
```
rolfbjarne, line 193: Maybe add a link to Apple's issue in the ignore message (or as a comment)?

rolfbjarne (approved):

tests/tom-swifty-test/SwiftReflector/PropertyWrapTests.cs
mandel-macaque:
```suggestion
		[Ignore ("Ignore due to  https://bugs.swift.org/browse/SR-13790")]
```

I think that is the correct one

mandel-macaque (approved):

spouliot (commented):
👍

Issue #448, opened 2020-09-21
When compiling ObjC swift types, we generate #nullable enable


stephen-hawley:
And when I say 'we' I mean btouch.

Likely we will need to adjust the arguments in this when we update the compiler version.

PR #497, opened 2020-11-03
Switch to csc

stephen-hawley:
This addresses the [issue](https://github.com/xamarin/binding-tools-for-swift/issues/448) with `#nullable` generated by btouch.

switch to `csc` from `mcs`, change `-r` arguments to include `.dll`.
Bob's your uncle.


spouliot (approved):
That's a bigger PR than I thought it would be 😸

chamons (approved):

PR #505, opened 2020-11-07
Getting Pipeline Build to work and some fixes to device tests

tj-devel709:
My apologies for this larger PR, in the future I will attempt to break PRs into smaller chunks for easier reviewing.

In this PR I have two major changes: 
- fixes to some of the device tests including naming conflicts and adding ignores until issues are fixed
- changes to get the build working on azure pipelines

Regarding dependencies:
- Moving to Xcode 12.0.1 since the VM does not have 12.0.0
- Moving Mono, Xamarin.iOS, and Xamarin.mac to versions supported by Catalina
- New hash in SwiftVersion.mk points to new commit in Binding-Tools-For-Swift-Reflector that replaces a branch that does not exist anymore

Big changes for the build:
- Disabling the 'Run Tests' part of the pipeline since there is a known issue of System.DllNotFoundException causing a lot of the tests to fail. Also because I turn off 'Run Tests', I believe we can remove turning off the tests in the makefile here: https://github.com/xamarin/binding-tools-for-swift/pull/505/files#diff-19433eb87fb3c276499be16f6e3c65a5e21556d88067eadac81d9783bdf5ca70L13
- Changing a header file to fix an issue with Catalina that causes clang to look for math.h in the wrong header file.

Question for @stephen-hawley or anyone else who may know: I used "#if _MAC_TS_TEST_" in some tests such as the one linked below that had a few tests that should only be run on mac. I know from another occasion, adding this was not necessary when the entire test file was to be run on mac. What is the appropriate way to handle this situation?
 https://github.com/xamarin/binding-tools-for-swift/pull/505/files#diff-500590aefd8e4bf3f98ac37b1174e42e52e97947862e47658c04c709c643ca45R107

### DIFF
--- a/Make.config
+++ b/Make.config
@@ -32,9 +32,9 @@ SOM_PACKAGE_VERSION=0.2.$(SOM_PACKAGE_VERSION_REV).$(SOM_COMMIT_DISTANCE)
 
 # The Xcode we need.
 # Use the path as used in the Azure Pipelines vm images.
-XCODE_VERSION=12.0
-XCODE_URL=http://xamarin-storage/bot-provisioning/xcodes/Xcode_12.0.0.xip
-XCODE_DEVELOPER_ROOT=/Applications/Xcode_12.0.0.app/Contents/Developer
+XCODE_VERSION=12.0.1
+XCODE_URL=http://xamarin-storage/bot-provisioning/xcodes/Xcode_12.0.1.xip
+XCODE_DEVELOPER_ROOT=/Applications/Xcode_12.0.1.app/Contents/Developer
 
 # Minimum Mono version we need
 MIN_MONO_VERSION=6.4.0.0
--- a/Make.config
+++ b/Make.config
@@ -42,9 +42,9 @@ MAX_MONO_VERSION=6.12.0.90
 MIN_MONO_URL=https://download.mono-project.com/archive/6.12.0/macos-10-universal/MonoFramework-MDK-6.12.0.90.macos10.xamarin.universal.pkg
 
 # XI version we need
-MIN_XI_VERSION=14.2.0.12
-MAX_XI_VERSION=14.2.0.12
-MIN_XI_URL=https://download.visualstudio.microsoft.com/download/pr/7b60a920-c8b1-4798-b660-ae1a7294eb6d/bbdc2a9c6705520fd0a6d04f71e5ed3e/xamarin.ios-14.2.0.12.pkg
+MIN_XI_VERSION=13.18.2.1
+MAX_XI_VERSION=13.18.2.1
+MIN_XI_URL=https://download.visualstudio.microsoft.com/download/pr/68ffa29a-6a5b-41f7-af7b-506ddcf4bbfc/35159cac3be1910e87309c0094a8ec8a/xamarin.ios-13.18.2.1.pkg
 
 # XM version we need
 MIN_XM_VERSION=6.20.2.2
@@ -52,6 +52,6 @@ MAX_XM_VERSION=6.20.2.2
 MIN_XM_URL=https://download.visualstudio.microsoft.com/download/pr/b089be2f-932a-40ab-904b-b626f9e6427b/6aad9f3ea4fbfb92ce267e0f60b34797/xamarin.mac-6.20.2.2.pkg
 
 # Minimum macOS version for building
-MIN_MACOS_BUILD_VERSION=10.15.0
+MIN_MACOS_BUILD_VERSION=10.15.4
 
 MIN_CMAKE_VERSION=3.18.4
--- a/Make.config
+++ b/Make.config
@@ -42,9 +42,9 @@ MAX_MONO_VERSION=6.12.0.90
 MIN_MONO_URL=https://download.mono-project.com/archive/6.12.0/macos-10-universal/MonoFramework-MDK-6.12.0.90.macos10.xamarin.universal.pkg
 
 # XI version we need
-MIN_XI_VERSION=13.18.2.1
-MAX_XI_VERSION=13.18.2.1
-MIN_XI_URL=https://download.visualstudio.microsoft.com/download/pr/68ffa29a-6a5b-41f7-af7b-506ddcf4bbfc/35159cac3be1910e87309c0094a8ec8a/xamarin.ios-13.18.2.1.pkg
+MIN_XI_VERSION=14.2.0.12
+MAX_XI_VERSION=14.2.0.12
+MIN_XI_URL=https://download.visualstudio.microsoft.com/download/pr/7b60a920-c8b1-4798-b660-ae1a7294eb6d/bbdc2a9c6705520fd0a6d04f71e5ed3e/xamarin.ios-14.2.0.12.pkg
 
 # XM version we need
 MIN_XM_VERSION=6.20.2.2
--- a/Make.config
+++ b/Make.config
@@ -42,9 +42,9 @@ MAX_MONO_VERSION=6.12.0.90
 MIN_MONO_URL=https://download.mono-project.com/archive/6.12.0/macos-10-universal/MonoFramework-MDK-6.12.0.90.macos10.xamarin.universal.pkg
 
 # XI version we need
-MIN_XI_VERSION=14.2.0.12
-MAX_XI_VERSION=14.2.0.12
-MIN_XI_URL=https://download.visualstudio.microsoft.com/download/pr/7b60a920-c8b1-4798-b660-ae1a7294eb6d/bbdc2a9c6705520fd0a6d04f71e5ed3e/xamarin.ios-14.2.0.12.pkg
+MIN_XI_VERSION=13.18.2.1
+MAX_XI_VERSION=13.18.2.1
+MIN_XI_URL=https://download.visualstudio.microsoft.com/download/pr/68ffa29a-6a5b-41f7-af7b-506ddcf4bbfc/35159cac3be1910e87309c0094a8ec8a/xamarin.ios-13.18.2.1.pkg
 
 # XM version we need
 MIN_XM_VERSION=6.20.2.2
--- a/Make.config
+++ b/Make.config
@@ -52,6 +52,6 @@ MAX_XM_VERSION=6.20.2.2
 MIN_XM_URL=https://download.visualstudio.microsoft.com/download/pr/b089be2f-932a-40ab-904b-b626f9e6427b/6aad9f3ea4fbfb92ce267e0f60b34797/xamarin.mac-6.20.2.2.pkg
 
 # Minimum macOS version for building
-MIN_MACOS_BUILD_VERSION=10.15.4
+MIN_MACOS_BUILD_VERSION=10.15.0
 
 MIN_CMAKE_VERSION=3.18.4
--- a/Make.config
+++ b/Make.config
@@ -37,17 +37,17 @@ XCODE_URL=http://xamarin-storage/bot-provisioning/xcodes/Xcode_12.0.1.xip
 XCODE_DEVELOPER_ROOT=/Applications/Xcode_12.0.1.app/Contents/Developer
 
 # Minimum Mono version we need
-MIN_MONO_VERSION=6.4.0.0
-MAX_MONO_VERSION=6.4.0.999
-MIN_MONO_URL=https://download.visualstudio.microsoft.com/download/pr/5b7dcb51-3035-46f7-a8cb-efe3a1da351c/dcba976cd3257636b6b2828575d29d3c/monoframework-mdk-6.4.0.208.macos10.xamarin.universal.pkg
+MIN_MONO_VERSION=6.12.0.90
+MAX_MONO_VERSION=6.12.0.90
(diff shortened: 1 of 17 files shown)


tests/tom-swifty-test/SwiftReflector/ProtocolTests.cs
```diff
@@ -104,6 +104,8 @@ void WrapSingleSubscriptGetOnly (string type, string csType, string csReplacemen
 			TestRunning.TestAndExecute (swiftCode, callingCode, expected, testName : $"WrapSingleSubscriptGetOnly{type}", otherClass : overCS);
 		}
 
+#if _MAC_TS_TEST_
```
stephen-hawley, line 107: `_MAC_TS_TEST_ ` doesn't make sense here.  It should only be used in generated code, not in an actual test. I want to be clear that my tone here might sound curt, but I want to be specific. It's NBD here, but it doesn't do what you think it does.

TL;DR - I want to make a test that runs only on a specific platform. How do I do that? There's an argument for that. When you call `TestAndExecute` there is an optional argument, `platform` (IIRC) which you can use to specify "I want this to run on....". If you don't specify it, the assumption is that it will run on all platforms upon which we run tests (currently macOS and iOS, but in the future...).

What is `_MAC_TS_TEST_` for then? A lot of it comes down to "thanks, swift, you're the best!". There are some tests that are written that do printing from swift (DON'T DO THIS). This turns out to be a bad idea because the (1) swift output and mono output are buffered differently and the ordering can be surprising and not always consistent. (2) I've yet to succeed in getting swift output redirected on iOS transparent to the swift code. (3) sometimes the set up and tear down for the actual test has things that only make sense on macOS.
Because of this, test generation code knows that it needs to make blocks.
For a typical unit test, the process is:
1. write swift code and compile it
2. creating bindings for the swift code and compile the swift parts
3. write C# code to invoke the binding in some meaningful way and generate output
4. compile the C# code and execute it
5. collect the output and see if it matches the expected `Assert.AreEqual (output, expected);`

If you look in `CaptureSwiftOutputPostlude` in `TestRunning.cs` you can see where `_MAC_TS_TEST_` gets injected into the code. If you look in `Compiler.cs` you can see where `_MAC_TS_TEST_` gets set as a compiler option (`-d _MAC_TS_TEST`).

If you put this around the actual `[Test]` block, what you're doing is commenting out the test so it never gets run.

Lessons learned for the future:
1. Testing binding is hard and you should think of this whole thing (gestures vaguely) as a dancing bear. It's not that it dances well that's impressive, it's that it dances at all.
2. Avoid writing unit tests that depend on output in swift. We can do it, but it's a serious pain and every time you run a unit test with swift output, a pixie gets set on fire.
3. If you can come up with a better way to unit test this (runs faster, easier to debug, etc), I am all ears. This is not defensive - I would seriously love an improvement to this process. I have a few ideas, but they're all trade-offs (faster tests, but harder to debug or faster tests but unrealistic testing scenario).

stephen-hawley (changes requested):

Make.config
```diff
@@ -32,26 +32,26 @@ SOM_PACKAGE_VERSION=0.2.$(SOM_PACKAGE_VERSION_REV).$(SOM_COMMIT_DISTANCE)
...
+MIN_MONO_VERSION=6.12.0.90
+MAX_MONO_VERSION=6.12.0.90
+MIN_MONO_URL=https://download.mono-project.com/archive/6.12.0/macos-10-universal/MonoFramework-MDK-6.12.0.90.macos10.xamarin.universal.pkg
 
 # XI version we need
-MIN_XI_VERSION=12.0.0.15
-MAX_XI_VERSION=12.0.0.15
-MIN_XI_URL=https://download.visualstudio.microsoft.com/download/pr/28396824-ecde-429e-9693-0da9382e1474/debd3c099bf5a516b964675c2628e601/xamarin.ios-12.0.0.15.pkg
+MIN_XI_VERSION=13.18.2.1
+MAX_XI_VERSION=13.18.2.1
```
spouliot, line 46: You might run off into some errors (like `mlauch`) since 13.18 was released before Xcode 12.
What's blocking you from using XI 14.x ?

Make.config
```diff
@@ -32,26 +32,26 @@ SOM_PACKAGE_VERSION=0.2.$(SOM_PACKAGE_VERSION_REV).$(SOM_COMMIT_DISTANCE)
...
-MIN_XM_VERSION=5.0.0.0
-MAX_XM_VERSION=5.0.0.0
-MIN_XM_URL=https://download.visualstudio.microsoft.com/download/pr/3849ffa9-0763-4727-8ae1-d67da9a6e60c/5c02d563ef9568fa79ba6dcf0cfeabd0/xamarin.mac-5.0.0.0.pkg
+MIN_XM_VERSION=6.20.2.2
+MAX_XM_VERSION=6.20.2.2
+MIN_XM_URL=https://download.visualstudio.microsoft.com/download/pr/b089be2f-932a-40ab-904b-b626f9e6427b/6aad9f3ea4fbfb92ce267e0f60b34797/xamarin.mac-6.20.2.2.pkg
 
 # Minimum macOS version for building
-MIN_MACOS_BUILD_VERSION=10.14.3
+MIN_MACOS_BUILD_VERSION=10.15.0
```
spouliot, line 55: Xcode 12.0.1 requires 10.15.4.

https://developer.apple.com/documentation/xcode-release-notes/xcode-12_0_1-release-notes

jenkins/build-swift.sh
```diff
@@ -105,6 +105,10 @@ if test -d swift; then
 	cd swift
 fi
 
+echo "Reinstalling six"
+pip uninstall six
+pip install six
+
```
spouliot, line 111: why the dance ?
if needed please add a comment

tests/tom-swifty-test/SwiftReflector/MetatypeTests.cs
```diff
@@ -172,6 +172,7 @@ public static void Main (string[] args)
 
 				string output = Compiler.RunWithMono (Path.Combine (temp.DirectoryPath, "TestIt.exe"), temp.DirectoryPath, platform: PlatformName.macOS);
 				Assert.AreEqual (expected, output);
+				var typeBasedClassName = typeName.Replace('.', '_');
```
spouliot, line 175: minor/style: space before `(`

tests/tom-swifty-test/SwiftReflector/MetatypeTests.cs
```diff
@@ -229,6 +230,7 @@ public static void Main (string[] args)
 
 				var output = TestRunning.Execute (temp.DirectoryPath, "TestIt.exe", PlatformName.macOS);
 				Assert.AreEqual (expected, output);
+				var typeBasedClassName = typeName.Replace('.', '_');
```
spouliot, line 233: same (style)
maybe you should add (if not already there) the `.editorconfig` file to the repo ?

whitneyschmidt (approved):

rolfbjarne (changes requested):

stephen-hawley (approved):

jenkins/build-swift.sh
```diff
@@ -115,6 +118,10 @@ git reset --hard "$SWIFT_HASH"
 echo "Updating swift dependencies..."
 ./utils/update-checkout --clone --skip-repository swift -j 1 --scheme "$SWIFT_SCHEME"
 
+echo "Pointing to different math.h for Clang/Catalina issue"
+NewMathHLocation='#include \"\.\/math\.h\"'
+sed -i '' -e "s/#include <math.h>/${NewMathHLocation}/" /Applications/Xcode_12.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/cmath
```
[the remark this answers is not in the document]
tj-devel709: Here is a link to the pipelines error:

https://dev.azure.com/xamarin/public/_build/results?buildId=29456&view=logs&j=b944efd9-1a3a-54bf-926f-a01c570708bc&t=60778944-7c7a-5fea-4140-c14bb1e329d7

Make.config
tj-devel709: When using XI 14.x, I get errors like this "ld: warning: object file (bin/Debug/mac/x86_64/ofiles/registeraccess.o) was built for newer **macOS** version (10.15) than being linked (10.9)".

I was having trouble tracking down this error. Steve told me what versions he was using and succeeding with so I tried his versions (which seemed compatible with macOS 10.15) so I transitioned to his XI 13.18.2.1. 
However, after you mentioned to bump MIN_MACOS_BUILD_VERSION to 10.15.4, that error up there changes to "ld: warning: object file (bin/Debug/appletv/arm64/ofiles/registeraccess.o) was built for newer **tvOS** version (14.0) than being linked (10.2)" 

Now I feel like there is just a tvOS version being set somewhere, but I have not seen it yet.

jenkins/build-swift.sh
tj-devel709: Okay I removed the uninstall because it is unnecessary and provided a comment on why I am installing six. A few lines below, './utils/build-script' is called and gets an error saying it cannot find six.

jenkins/build-swift.sh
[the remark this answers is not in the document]
rolfbjarne: Oh, this is while building swift. I googled a bit, and found this: https://bugs.swift.org/browse/SR-12726.

Might be fixed by cherry-picking this into the swift branch we're using: https://github.com/apple/swift/commit/4d3bd22c4e281ba0c1f7685261cce1e5fb6981af

I think that would be the easiest solution, but the bug has a few other suggestions (uninstall the command line tools, or uninstall pkg-config), which could also work.

Make.config
rolfbjarne: You should be able to run `make V=1` in swiftglue/ and get a more verbose log (and one where everything is run serialized, so the warning will apply to the last command shown).

In any case, this is probably because of a missing compiler flag somewhere (my guess would be `-mmacosx-version-min=10.9`, and the equivalent for the other platforms), but I might be wrong here.

tj-devel709: Oops, my apologies. I was looking at the warning and not the error! The error has to do with modules not being found in this section "SWIFTC [iphone/x86_64/Release]  XamGlue" at https://dev.azure.com/xamarin/public/_build/results?buildId=30396&view=logs&j=b944efd9-1a3a-54bf-926f-a01c570708bc&t=21e7d97e-20fa-52e2-1c14-34bcaa2b17dd&l=259

This happens regardless of the minimum set macOS version and changing the XI from 13.18.2.1 to 14.2.0.12 seems to trigger this. This happens to me locally as well.

tests/tom-swifty-test/SwiftReflector/MetatypeTests.cs
tj-devel709: Will add this on my todos :)

Make.config
rolfbjarne: Wild guess: you might have to add `MediaSetup` to the list of frameworks here: https://github.com/xamarin/binding-tools-for-swift/blob/e54898827eaf58c9d3e4c8b0c90fbe27e65f5a81/common.mk#L137

tj-devel709: Oh man.. Thank you haha I'll give that a try!

tj-devel709: Sadly this did not fix this issue. Perhaps I will mark this as an issue and attempt to fix this in a later PR since there are other fixes in this PR that are blocking @chamons. Is this okay?

rolfbjarne (approved):

tj-devel709:
Issue with XI version filed here: https://github.com/xamarin/binding-tools-for-swift/issues/506

jenkins/build-swift.sh
mandel-macaque: You are going to have a few issue wit this:

* There is no guarantee that pip will be present in the machine. i have seen that the 10.15 machines do not have pip unless you specify it in the image.
* Installing python modules in the machine for a build.. is not ideal. You are using a vm, so is not an issue BUT you are also doing it on machines. Something to consider.

jenkins/system-dependencies.sh
```diff
@@ -404,7 +404,7 @@ function check_specific_xcode () {
 
 	if ! test -d "$XCODE_DEVELOPER_ROOT"; then
 		if ! test -z "$PROVISION_XCODE"; then
-			if ! test -z "$JENKINS_URL"; then
+			if ! test -z "$SYSTEM.ACCESSTOKEN"; then
```
mandel-macaque, line 407: Weird change, what is the diff?

mandel-macaque (approved):
With some comments. But no blockers.

jenkins/system-dependencies.sh
tj-devel709: This was something that Connor suggested I do. There was an error in provision dependencies and Connor said that the test was returning false because it's currently not running in a Jenkins instance and to change to check for DevOps variable

jenkins/build-swift.sh
tj-devel709: Thanks, I will add this to my radar and fix in next PR!

PR #509, opened 2020-11-17
Let's get unit tests going on the CLI

stephen-hawley:
Problem: Unit tests were passing in the IDE, but failing in the CLI.
Tracking it down, the issue was due to the DYLD_LIBRARY_PATH being set in the test Makefile in such a way that included the output libraries built alongside the reflector. No swift code that we use is compiled against these libraries and they are incompatible with the system libraries. The solution was to replace the reflector library path with `/usr/lib/swift`.

In the process, I added a ton (metric) of debugging information. This includes:
1. adding in an optional method to the test rig and a method that dumps all the directories in `DYLD_LIBRARY_PATH` and their contents
2. Adding in a SoleMetadataAccessor for types
3. Explicitly searching through `DYLD_LIBRARY_PATH` when doing DlOpen and improving the output on failure
4. Adding code that can get a metadata accessor function by its symbol instead of just using a direct metadata symbol
5. Added `CSShortCircuit` to Dynamo to allow us to write code to include `continue` or `break`

When the code was finally running, one test failed: `SwiftArrayTests.Constructor_Capacity`, which included an range check test for the flavor of constructor that takes an `nint`. This test failed because it was being called in the form `new SwiftArray<long>((long)-1);` Using `mcs` previously, this went to the intended constructor. Now that the code is compiling with `csc` it was now going to `SwiftArray<T>(params T [] items)`, which does not fail the range check. Changed the test to be `new SwiftArray<int>((long)-1)` which now range checks as expected.

### DIFF
--- a/SwiftRuntimeLibrary/SwiftMarshal/DynamicLib.cs
+++ b/SwiftRuntimeLibrary/SwiftMarshal/DynamicLib.cs
@@ -99,10 +99,8 @@ namespace SwiftRuntimeLibrary.SwiftMarshal {
 		{
 			var path = Environment.GetEnvironmentVariable ("DYLD_LIBRARY_PATH");
 			if (path == null)
-				yield break;
-			foreach (var subPart in path.Split (':')) {
-				yield return subPart;
-			}
+				return new string [0];
+			return path.Split (':');
 		}
 
 		public string FileName { get; private set; }
--- a/SwiftRuntimeLibrary/SwiftMarshal/DynamicLib.cs
+++ b/SwiftRuntimeLibrary/SwiftMarshal/DynamicLib.cs
@@ -2,7 +2,9 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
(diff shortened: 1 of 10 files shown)


tj-devel709 (approved):

SwiftRuntimeLibrary/SwiftMarshal/DynamicLib.cs
```diff
@@ -86,6 +95,16 @@ public DynamicLib (string fileName, DLOpenMode flags)
...
 		}
 
+		IEnumerable<string> LibraryPaths ()
+		{
+			var path = Environment.GetEnvironmentVariable ("DYLD_LIBRARY_PATH");
+			if (path == null)
+				yield break;
+			foreach (var subPart in path.Split (':')) {
+				yield return subPart;
+			}
```
rolfbjarne, line 105: I think something like this should work:

```suggestion
			if (path == null)
				return Array.Empty<string> ();
			return path.Split (':');
```

rolfbjarne (approved):

spouliot (approved):

PR #536, opened 2021-02-03
[CI] Ensure that we do not use relative paths.

mandel-macaque:
Cheking out two or more repositories using the yaml results in a diff
behaviour regarding the value of the currenct working directory. While
when we only have one repo, the variable points to the checkout, if we
have more, it points to the dir with ALL the repos.

There are a number of scripts that are making the wrong assumption about
their location, resulting in 'strange' issues (not really strage, the
path does not exist because you are not longer where you think you are).

We fix this by:

* Allways use absolute paths via the sources dir variable.
* Add a step that will set up the root of the two repos we are
  interested to be used in the scripts.

This should make the CI less surprising in the future.

There is a second needed fix. A status is related to the hash of the
commit, VSTS is merforming a checkout and a merge, which results in a
new commit that does not exist in the remote. With that in mind we have
the following:

1. $(Build.SourceVersion) - points to the hash of the merge.
2. $(System.PullRequest.SourceCommitId)) - points to the hash of the
   last commit of the PR that triggered the build.

Because we want to set a status, the hash has to be known by the remote,
ergo we are interested in 2.


tj-devel709 (approved):
Ah thanks for fixing my mistakes!! 😄

rolfbjarne (approved):

dalexsoto (approved):

Issue #519, opened 2021-01-21
SwiftInterfaceReflector doesn't do unavailable/deprecated


stephen-hawley:
To be fair, the swift reflector doesn't handle these correctly, but it generates something.
The correct way to handle these is to make `<Attribute>` elements for anything that can have attributes so they can get passed through.
We can implement this (they'll get ignored in the XML for now) and use them for figuring out unavailable/deprecated.

PR #548, opened 2021-02-12
Added isUnavailable and isDeprecated

stephen-hawley:
Added calculations for `isUnavailable` and `isDeprecated` fixing issue [519](https://github.com/xamarin/binding-tools-for-swift/issues/519).

I handled `isUnavailable` and `isDeprecated` wrong in the swift reflector and it should go away, but we have code that depends on it, so here we are. In the reflector, these values are taken from the compiler and there is apparently a lot more work that needs to have to figure them out and they only represent a snapshot of what is happening for the compiler settings currently.

In reality, there is a lot more information. The best explanation I found of the `@available` attribute is [here](https://nshipster.com/available/), but I'll sum up.

There are two versions of `@available`, longhand and shorthand. Shorthand looks like this:
```
@available(platform vers[, platform vers]*, *)
```
So if you see this:
```
@available(macOS 10.0, iOS 12.2.1, *)
```
it is exactly equivalent to this:
```
@available(macOS, introduced: 10.0)
@available(iOS, introduces: 12.2.1)
```

Longhand looks like this:
```
@available(platform[, keyword: arg][, unavailable])
```
where `keyword` is one of `introduced`, `deprecated`, `obsoleted`, `message` `renamed`. The last two take a string. The rest take an optional version number.

Deprecated is pretty easy. If it's longhand and the OS matches and it contains deprecated and the version number isn't present or the version number is less than or equal to the OS version setting.

Unavailable is trickier. If the attribute contains `unavailable` and the OS matches then it's unavailable or it's not available.
It's available if it was introduced before the OS version specified, but we have to check for both shorthand and longhand.

A lot of this code does a lot of digging in and parsing and making assumptions about the layout of things and frankly it gives me the screaming heebie-jeebies. One such area is getting a `Version` from the attribute which may or may not be present. There are assumptions such as when in shorthand, you will always see a comma after the version since the last element of shorthand has to be ", *". For longhand, you will see either `key: version,` or `key, otherkey` or `key` (where this is the last element of the list). So we look for the index of the first `:` after the key. If there we hit a comma first or reach the end of tags, then we use Version 0.0, which means since the beginning of time which is good enough.

In the process, I found an error in two unit tests with ObjC selectors and the caseness of the `Public` `Private` labels and they were minor so I just fixed them here since they were preventing the tests that I wrote from passing.

### DIFF
--- a/SwiftReflector/SwiftInterfaceReflector/SwiftInterfaceReflector.cs
+++ b/SwiftReflector/SwiftInterfaceReflector/SwiftInterfaceReflector.cs
@@ -37,6 +37,10 @@ namespace SwiftReflector.SwiftInterfaceReflector {
 		internal const string kPrivate = "private";
 		internal const string kInternal = "internal";
 		internal const string kOpen = "open";
+		internal const string kPublicCap = "Public";
+		internal const string kPrivateCap = "Private";
+		internal const string kInternalCap = "Internal";
+		internal const string kOpenCap = "Open";
 		internal const string kFilePrivate = "fileprivate";
 		internal const string kStatic = "static";
 		internal const string kIsStatic = "isStatic";
@@ -108,6 +112,11 @@ namespace SwiftReflector.SwiftInterfaceReflector {
 		internal const string kSublist = "Sublist";
 		internal const string kValue = "Value";
 		internal const string kObjCSelector = "objcSelector";
+		internal const string kDeprecated = "deprecated";
+		internal const string kUnavailable = "unavailable";
+		internal const string kAvailable = "available";
+		internal const string kIntroduced = "introduced";
+		internal const string kObsoleted = "obsoleted";
 
 		Stack<XElement> currentElement = new Stack<XElement> ();
 		Version interfaceVersion;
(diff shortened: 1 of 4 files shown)


SwiftReflector/SwiftInterfaceReflector/SwiftInterfaceReflector.cs
```diff
@@ -1103,6 +1116,259 @@ List<XElement> GatherInheritance (Type_inheritance_clauseContext context, bool f
...
+			case "swift":
+				return true;
+			case "iOS":
+			case "iOSApplicationExtension":
+				return currentPlatform.StartsWith ("ios", StringComparison.Ordinal);
+			case "macOS":
+			case "macOSApplicationExtension":
+			case "macCatalyst":
+			case "macCatalystExtension":
+				return currentPlatform.StartsWith ("macos", StringComparison.Ordinal);
```
spouliot, line 1320: macOS and MacCatalyst are different platforms, the latter will use version numbers that match iOS since it's largely based on that code

spouliot (changes requested):

SwiftReflector/SwiftInterfaceReflector/SwiftInterfaceReflector.cs
stephen-hawley: OK - I'll open up an issue on that. I wasn't able to generate that tag from the compiler easily so I guessed (incorrectly).

stephen-hawley: And I addressed it - not so bad.

spouliot (approved):

dalexsoto (approved):

PR #550, opened 2021-02-16
Xml compare

stephen-hawley:
This is a tool I'm going to need for further testing.
It compares 2 `XDocument` trees by looking at the elements ND attributes.
It needs to be clear that this is **intentionally** a one-way comparison. This is because ultimately this will be used to compare two XML trees where the second is allowed (and will) have `XElement` elements that are not present in the first and may have attributes not in the first and that's acceptable.


spouliot (commented):
It's not complete - but a complete solution requires C14N and a lot of code (pain).

Is this for testing purpose only ? or is this exposed / used by customers to produce bindings ?

If the latter maybe adding a comment on the `XmlComparator` to ensure it's not re-used for other purposes ?

stephen-hawley:
> It's not complete - but a complete solution requires C14N and a lot of code (pain).
> 
> Is this for testing purpose only ? or is this exposed / used by customers to produce bindings ?
> 
> If the latter maybe adding a comment on the `XmlComparator` to ensure it's not re-used for other purposes ?

It is for testing only, which is why it lives in the test project only. I'll make a note and put in some comments.

spouliot (approved):

PR #563, opened 2021-02-17
Added the ability to select the reflector in tests

stephen-hawley:
Added the ability to select the reflector to be used in the tests.
Changed all the DynamicXmlTests to have a test case for each mode.
Ignored the failing cases (opened issues for them).

### DIFF
--- a/SwiftReflector/CustomSwiftCompiler.cs
+++ b/SwiftReflector/CustomSwiftCompiler.cs
@@ -167,6 +167,33 @@ namespace SwiftReflector {
 			return stm;
 		}
 
+		public XDocument ReflectToXDocument (IEnumerable<string> includeDirectories,
+								 IEnumerable<string> libraryDirectories, string extraArgs,
+		                                                 params string [] moduleNames)
+		{
+			var moduleNameAggregate = moduleNames.Aggregate ((s1, s2) => s1 + s2);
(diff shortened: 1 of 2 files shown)


SwiftReflector/CustomSwiftCompiler.cs
```diff
@@ -167,6 +167,33 @@ public Version GetCompilerVersion ()
...
+													     includeDirectories, CompilerInfo.Target);
+			string output = "";
+			try {
+				output = Reflect (locations.Select (loc => loc.DirectoryPath), libraryDirectories, pathName, extraArgs, moduleNames);
+			} finally {
+				locations.DisposeAll ();
+			}
+			ThrowOnCompilerVersionMismatch (output, moduleNames);
+			using (var stm = new FileStream (output, FileMode.Open, FileAccess.Read, FileShare.ReadWrite)) {
+				return XDocument.Load (stm);
```
spouliot, line 192: why not directly call `public static System.Xml.Linq.XDocument Load (string uri);` ?

spouliot (approved):

SwiftReflector/CustomSwiftCompiler.cs
stephen-hawley: Because I didn't want to assume that a file system path would work for a uri instead of a `file://`?

Issue #551, opened 2021-02-17
Reflector gets public names for args with no explicit public name wrong


stephen-hawley:
coming through as "_" should be "";

PR #568, opened 2021-02-19
Fixing mismatched parameter names

stephen-hawley:
Fix parameter naming, correcting issue [551](https://github.com/xamarin/binding-tools-for-swift/issues/551)

Oh, boy do I hate swift parameter naming.
Background:
swift lets you name parameters in 3 different ways:
```
a b: type
_ b: type
a: type
```
If there was no explicit public name (third case), I was putting in "_" which is wrong.
In my world, I keep the public names and private names like this per case:
1. "a" "b"
2. "" "b"
3. "a" "a"

Where it gets "funny" is for subscripts which have implicitly empty public argument names and for some reason (really, don't ask me), I want them to be "_". *facepalm*

This brings 4 tests online.

### DIFF
--- a/SwiftReflector/CustomSwiftCompiler.cs
+++ b/SwiftReflector/CustomSwiftCompiler.cs
@@ -188,7 +188,7 @@ namespace SwiftReflector {
 				locations.DisposeAll ();
 			}
 			ThrowOnCompilerVersionMismatch (output, moduleNames);
-			using (var stm = new FileStream (output, FileMode.Open, FileAccess.Read, FileShare.ReadWrite)) {
+			using (var stm = new FileStream (pathName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite)) {
 				return XDocument.Load (stm);
 			}
 		}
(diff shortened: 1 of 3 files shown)


SwiftReflector/CustomSwiftCompiler.cs
```diff
@@ -188,7 +188,7 @@ public Version GetCompilerVersion ()
 				locations.DisposeAll ();
 			}
 			ThrowOnCompilerVersionMismatch (output, moduleNames);
-			using (var stm = new FileStream (output, FileMode.Open, FileAccess.Read, FileShare.ReadWrite)) {
+			using (var stm = new FileStream (pathName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite)) {
```
spouliot, line 191: how is this related to the other changes ?

stephen-hawley: Because I used this code to debug it and it's using the the error string as a path name. It didn't work, so I fixed it so I could debug.

Issue #553, opened 2021-02-17
Xml Parsing causing an NRE


stephen-hawley:
FuncReturningIntThrows (and several others) are causing an NRE.

PR #572, opened 2021-02-22
make enums more complete, remove NRE

stephen-hawley:
Fixes an NRE in XML parsing of enums, issue [553](https://github.com/xamarin/binding-tools-for-swift/issues/553).

To look at this, there were two missing pieces in enums: the cases and the raw value type.

For the cases, there are two broad varieties: a raw value enum type and a union enum type, but both of these have the same structure: in the Enter event of each of these, grab the cases and bundle them up into an elements tag and add that to the current enum. Since they can come out of order, if the elements tag already exists in the parent enum, we just add in the contents.

Converting union enums is...funny. This is because the syntax is fairly broad:
`union_style_enum_case : enum_case_name tuple_type? ;`
So you can have a full tuple syntax which is cool and all, except that in most cases it's a single type, but it gets expressed as a single element tuple. In that case, we strip out the parentheses and if there is a label on the type, we strip that too.

The exciting part here is swift allows some really...unusual types as a single tuple including:
```
case foo (())
case bar (bar:())
case baz (((((())))))
case this_is_illegal() // syntax error
```
If the enum is a trivial enum, there will be a raw value as a type alias that we dig out and put into an attribute.

Previously ignored NRE tests pass.

### DIFF
--- a/SwiftReflector/SwiftInterfaceReflector/SwiftInterfaceReflector.cs
+++ b/SwiftReflector/SwiftInterfaceReflector/SwiftInterfaceReflector.cs
@@ -117,6 +117,11 @@ namespace SwiftReflector.SwiftInterfaceReflector {
 		internal const string kAvailable = "available";
 		internal const string kIntroduced = "introduced";
 		internal const string kObsoleted = "obsoleted";
+		internal const string kElements = "elements";
+		internal const string kElement = "element";
+		internal const string kIntValue = "intValue";
+		internal const string kRawType = "rawType";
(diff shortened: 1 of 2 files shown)


spouliot (approved):
> Previously ignored NRE tests pass.

Should not they be re-enabled ?

stephen-hawley:
Good catch!

dalexsoto (approved):

tj-devel709 (approved):

rolfbjarne (approved):

PR #588, opened 2021-03-09
Allow reflection to be done as a strategy.

stephen-hawley:
These changes allow for primary and secondary reflection strategies.

I've set the ordering to be compiler then parser since the parser is generating a few (hundred) errors right now.
I figured I would sort those out one class of error at a time before activating this.

### DIFF
--- a/SwiftReflector/CustomSwiftCompiler.cs
+++ b/SwiftReflector/CustomSwiftCompiler.cs
@@ -14,6 +14,8 @@ using System.Linq;
 using Dynamo;
 using System.Text.RegularExpressions;
 using ObjCRuntime;
+using SwiftReflector.TypeMapping;
+using SwiftReflector.SwiftInterfaceReflector;
 
 namespace SwiftReflector {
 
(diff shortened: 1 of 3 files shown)


SwiftReflector/CustomSwiftCompiler.cs
```diff
@@ -33,6 +35,40 @@ public SwiftCompilerOptions (string moduleName, string [] includeDirectories, st
...
+				if (TryLoadFile (file, into, "", ".xml", ".XML"))
+					return true;
+			}
+			return false;
+		}
+
+		bool TryLoadFile (string file, TypeDatabase into, params string [] extensions)
+		{
+			foreach (var ext in extensions) {
+				var candidate = file + ext;
```
spouliot, line 60: Path.ChangeExtension
https://docs.microsoft.com/en-us/dotnet/api/system.io.path.changeextension?view=net-5.0

spouliot (approved):

rolfbjarne (approved):

Issue #612, opened 2021-03-16
SwiftInterfaceParser is messing up optional argument names


stephen-hawley:
see `NewClassCompilerTests.TestAnonParamInMethod`

PR #613, opened 2021-03-16
No Underscores for Parameter Names

stephen-hawley:
Remove underscores as parameter names fixing issue [612](https://github.com/xamarin/binding-tools-for-swift/issues/612)


spouliot (approved):
The issue mention `NewClassCompilerTests.TestAnonParamInMethod`.
Should that be re-enabled ?

stephen-hawley:
Right now all tests are enabled, but none of they're being run with the compiler as the first line reflector. When I get them all green, I will change this to use the parser as first line, although I suppose at this point I could activate it as it I'm down to only 34 failures.

PR #634, opened 2021-03-25
[CI] Run tests after months not running them.

mandel-macaque:
This will probably 💣 💥

### DIFF
--- a/devops/automation/run-tests.sh
+++ b/devops/automation/run-tests.sh
@@ -1,5 +1,7 @@
 #!/bin/bash -e
 
+echo $SOM_PATH
+unset SOM_PATH
 cd "$(dirname "${BASH_SOURCE[0]}")/../.."
 #WORKSPACE=$(pwd)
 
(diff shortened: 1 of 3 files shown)


rolfbjarne (approved):

devops/automation/run-tests.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash -e
 
+echo $SOM_PATH
+unset SOM_PATH
```
tj-devel709, line 4: What is this doing?

tj-devel709:
Also I remember that when I started, I did a little work with device tests and a lot of them would run locally but they would not run in vsts because of naming conflicts. Just wanted to make sure this is not the issue that is happening now here. @mandel-macaque @stephen-hawley

devops/automation/run-tests.sh
mandel-macaque: I think it was stuff left behind.

mandel-macaque:
@tj-devel709 this issue I was fixing is that a commit you made in Nov disabled the tests on CI and have  been ran since then :/

Now we have failing tests that we did not know about.

stephen-hawley (approved):

Issue #641, opened 2021-04-02
SwiftInterfaceParser Should Ignore XamGlue


stephen-hawley:
SwiftInterface parser tries to load module information for XamGlue. It probably shouldn't since there really aren't any types in there that we care about.

It should be safe to skip this module entirely.

PR #644, opened 2021-04-05
fix for module loading

stephen-hawley:
Fixes two related issues in module loading fixing issue [641](https://github.com/xamarin/binding-tools-for-swift/issues/641)
The first is that we are instructed to try to load the module "XamGlue" which doesn't generally have any types in it with one exception: `EveryProtocol` which will only get seen as an extension. There are no type database entries for it, so loading always failed. I special cased that out.
I also put in a fallback in the type searching that would return `XamGlue.EveryProtocol` if the type was `EveryProtocol`. This is intentionally done __after__ the rest of the modules have been searched on the off-chance that someone else named a type `EveryProtocol` and to give that type priority.

There was also one other minor but related issue where if a module doesn't export any types of operators then the type database doesn't know anything about the module. I refactored the type database code to have an entry point that can assert the existence of a module even if it has no types in it. In the bottleneck where we add all the types to the type database, I make that assertion.

### DIFF
--- a/SwiftReflector/SwiftInterfaceReflector/SwiftInterfaceReflector.cs
+++ b/SwiftReflector/SwiftInterfaceReflector/SwiftInterfaceReflector.cs
@@ -1636,6 +1636,8 @@ namespace SwiftReflector.SwiftInterfaceReflector {
 		{
 			var failures = new StringBuilder ();
 			foreach (var module in importModules) {
+				if (module == "XamGlue")
+					continue;
 				if (!moduleLoader.Load (module, typeDatabase)) {
 					if (failures.Length > 0)
 						failures.Append (", ");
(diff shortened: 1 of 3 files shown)


SwiftReflector/SwiftInterfaceReflector/SwiftInterfaceReflector.cs
```diff
@@ -1899,13 +1901,15 @@ string ReplaceLocalName (string nonQualified)
 		}
 
 		string ReplaceGlobalName (string nonQualified)
-		{
+		{
```
tj-devel709, line 1904: Whitespace added by accident?

tj-devel709 (approved):

SwiftReflector/SwiftInterfaceReflector/SwiftInterfaceReflector.cs
stephen-hawley: 👍

dalexsoto (approved):

Issue #647, opened 2021-04-07
In macOS tests there is chatter in stderr


stephen-hawley:
check any of the NSObject tests or any test that compiles to a macOS app. There is a bit of chatter in stderr which is currently ignored. It has mostly to do with two copies of a library being available. This is a minor issue in general and is most likely due to having a redundant path element. Do we care? Probably not so much.

Issue #643, opened 2021-04-05
SwiftInterfaceParser is having trouble with ObjC tests


stephen-hawley:
It looks like there is other back channel chatter coming though on these tests.
Look to see if there is a linking issue that we need to consider.
```
2) Failed : SwiftReflector.GenericFunctionTests.WrapGenClassProtocolConstraintCorrectUsage
  Expected string length 2 but was 659. Strings differ at index 0.
  Expected: "1\n"
  But was:  "objc[63087]: Class AMSupportURLConnectionDelegate is implemen..."
  -----------^
  at tomwiftytest.TestRunning.TestAndExecute (System.String swiftCode, Dynamo.CodeElementCollection`1[T] callingCode, System.String expectedOutput, System.String testName, Dynamo.CSLang.CSClass otherClass, System.String skipReason, System.String iosExpectedOutput, SwiftReflector.PlatformName platform, SwiftReflector.UnicodeMapper unicodeMapper, System.Int32 expectedErrorCount, System.Action`1[T] postCompileCheck, System.String[] expectedOutputContains, System.Boolean enforceUTF8Encoding) [0x00346] in <a0e5c1328265423bbdb853ccd27baa3b>:0 
  at SwiftReflector.GenericFunctionTests.WrapGenClassProtocolConstraintCorrectUsage () [0x0016d] in <a0e5c1328265423bbdb853ccd27baa3b>:0 
```

PR #648, opened 2021-04-07
fix up mmp issues.

stephen-hawley:
Sort out mmp problem, fixing issue [643](https://github.com/xamarin/binding-tools-for-swift/issues/643)

This is partly an exercise in finding an issue and partly an issue in making life better for future engineer.

In this issue, a number of tests were failing running `mmp` which was due to `pkg-config` failing.
It took a while to isolate the issue with pkg-config. This took a while because running the mmp command from a shell worked fine. To manage this, I injected some code to print out all the environment variables and then started up a shell with an empty environment and ran a script with the code. This duplicated the error.
From there it was a matter of finding either enabling variable(s) in my usual shell or disabling variables(s) in the test runtime.

It turned out to be test runtime environment.

I put code in `ExecAndCollect` to dump the environment in addition to the command. This means that the entire output can go into a script if something like this turns up in the future.

In addition, I noticed that there was some chatter in stderr that was interfering with test output. I opened an issue on that [here](https://github.com/xamarin/binding-tools-for-swift/issues/647) and made the decision to only include stderr when the return from executing something was non-0.

There are two remaining failing tests which are ignored with issues opened on them.

### DIFF
--- a/tests/tom-swifty-test/TestRunning.cs
+++ b/tests/tom-swifty-test/TestRunning.cs
@@ -686,7 +686,14 @@ public static class Console {
 					var output = new StringBuilder ();
 					var responseFile = Path.Combine (workingDirectory, name + ".rsp");
 					File.WriteAllText (responseFile, mmp.ToString ());
-					var rv = ExecAndCollect.RunCommand ("/Library/Frameworks/Xamarin.Mac.framework/Versions/Current/bin/mmp", "--link_flags=-headerpad_max_install_names " + StringUtils.Quote ($"@{responseFile}"), output: output, verbose: true);
+					// The test environment is coming is with PKG_CONFIG_LIBDIR=""
+					// which causes macOS tests to fail. This removes the environment
+					// variable and lets the tests pass.
+					var env = new Dictionary<string, string> () {
+						{ "PKG_CONFIG_LIBDIR", null }
+					};
+					var rv = ExecAndCollect.RunCommand ("/Library/Frameworks/Xamarin.Mac.framework/Versions/Current/bin/mmp", "--link_flags=-headerpad_max_install_names " + StringUtils.Quote ($"@{responseFile}"),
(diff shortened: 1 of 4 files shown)


tests/tom-swifty-test/TestRunning.cs
```diff
@@ -686,7 +686,14 @@ public static string Execute (string workingDirectory, string executable, Platfo
 					var output = new StringBuilder ();
 					var responseFile = Path.Combine (workingDirectory, name + ".rsp");
 					File.WriteAllText (responseFile, mmp.ToString ());
-					var rv = ExecAndCollect.RunCommand ("/Library/Frameworks/Xamarin.Mac.framework/Versions/Current/bin/mmp", "--link_flags=-headerpad_max_install_names " + StringUtils.Quote ($"@{responseFile}"), output: output, verbose: true);
+					// The test environment is coming is with PKG_CONFIG_LIBDIR=""
```
mandel-macaque, line 689: do we know who is setting that? We could remove it

mandel-macaque (approved):

tests/tom-swifty-test/TestRunning.cs
stephen-hawley: No, no we don't.

tests/tom-swifty-test/XmlReflectionTests/SwiftInterfaceParserTests.cs
```diff
@@ -268,6 +268,7 @@ public class Bar : Foo, Nifty {
 		}
 
 		[Test]
+		[Ignore ("Throwing wrong exception ")]
```
tj-devel709, line 271:
```suggestion
		[Ignore ("Throwing wrong exception")]
```
Just a little extra space here!

tj-devel709 (approved):

PR #649, opened 2021-04-09
unstumble device builds

stephen-hawley:
Ensure that the right compiler flags are set
Ensure that swift interface files are copied
Deduplicate test code
Don't device test unit tests that conflict.
Took devices tests back off line because of a signing issue, [here](https://github.com/xamarin/binding-tools-for-swift/issues/650).


tj-devel709 (approved):

mandel-macaque (approved):
Looks good, but fails tests, are those expected?

stephen-hawley:
That's a good question. I can't reproduce that particular failure, so I maid the disposing predictable.

Issue #656, opened 2021-04-20
SwiftInterface parser failing to load modules


stephen-hawley:
RegisterAccess - ProtoWitnessTests
Cocoa - NSObjectTests

PR #662, opened 2021-04-21
Handle unknown module types and global aliases

stephen-hawley:
This fixes issues with module references that are likely wrong fixing issue [656](https://github.com/xamarin/binding-tools-for-swift/issues/656).

This is the type of issue where peeling one layer reveals and opportunity for heavy-handed refactoring. Fortunately we have the infrastructure in place to handle all of this.

There were 3 actual errors:
1. `Cocoa` doesn't really exist as a namespace AFAICT
2. `RegisterAccess` isn't a namespace that will be used for binding
3. `Foundation.NSRect` doesn't exist as a real type, it's an alias

These are all handled in different ways.
The first we blindly map `Cocoa` to `AppKit`. Should this be something else? Don't know. PRs welcome.
The second is not a namespace that will be used for binding and it exports no types, so we ignore it.
The third. Well.
At present, when a module is read in via reflection, there may be type databases associated with the module. Those get aggregated and any types that match one of those aliases will automagically get mapped to the correct type. This is great, but after the module is read in, the type aliases get thrown away.

No longer. Now the type database for a module will include all its type aliases for future use.
So now the formal process is that when a module is reflected, all of its type aliases and all the type aliases in the type database will get used to remap aliases.

The refactoring for this means:
- having a `TypeAliases` member on the `ModuleDatabase` type
- Adding code to write out the type aliases for a `ModuleDatabase`
- Adding code to read in the type aliases for a `ModuleDatabase`
- Adding code to merge all type databases aliases into the folder
- Refactoring all the code to read in XML reflection to include a `TypeDatabase` so we can get at the aliases

tl;dr - type aliases are now first class entities and handled formally in the process of reflecting on a module.

### DIFF
--- a/SwiftReflector/CustomSwiftCompiler.cs
+++ b/SwiftReflector/CustomSwiftCompiler.cs
@@ -44,6 +44,8 @@ namespace SwiftReflector {
 
 		public bool Load (string moduleName, TypeDatabase into)
 		{
+			if (moduleName == "Cocoa")
+				moduleName = "AppKit";
 			if (into.ModuleNames.Contains (moduleName))
 				return true;
 			foreach (var location in locations) {
@@ -249,7 +251,7 @@ namespace SwiftReflector {
 			locations.AddRange (includeDirectories);
 
 			ReflectWithStrategies (locations, libraryDirectories, pathName, extraArgs, moduleNames);
-			return Reflector.FromXmlFile (pathName);
+			return Reflector.FromXmlFile (pathName, ReflectionTypeDatabase);
 		}
 
 		public void ReflectWithStrategies (IEnumerable<string> includeDirectories, IEnumerable<string> libraryDirectories,
--- a/SwiftReflector/TypeMapping/TypeDatabase.cs
+++ b/SwiftReflector/TypeMapping/TypeDatabase.cs
@@ -176,6 +176,14 @@ namespace SwiftReflector.TypeMapping {
 			return Enumerable.Empty<OperatorDeclaration> ();
 		}
 
+		public IEnumerable<TypeAliasDeclaration> TypeAliasesForModule (string moduleName)
+		{
+			var module = EntityCollection (moduleName);
+			if (module != null)
+				return module.TypeAliases;
+			return Enumerable.Empty<TypeAliasDeclaration> ();
+		}
+
 		public void Write (string file, IEnumerable<string> modules)
 		{
 			using (FileStream stm = new FileStream (Exceptions.ThrowOnNull (file, nameof(file)), FileMode.Create)) {
@@ -196,6 +204,9 @@ namespace SwiftReflector.TypeMapping {
 				}
 				foreach (var op in OperatorsForModule (modName)) {
 					yield return op.ToXElement ();
+				}
+				foreach (var alias in TypeAliasesForModule (modName)) {
+					yield return alias.ToXElement ();
 				}
 			}
 		}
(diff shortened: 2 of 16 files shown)


SwiftReflector/CustomSwiftCompiler.cs
```diff
@@ -44,6 +44,8 @@ public FileSystemModuleLoader (IEnumerable<string> locations)
 
 		public bool Load (string moduleName, TypeDatabase into)
 		{
+			if (moduleName == "Cocoa")
+				moduleName = "AppKit";
```
spouliot, line 48: `AppKit` is macOS specific
Do you get `Cocoa` for iOS or other OS ?
What are those symbols that refers to `Cocoa` ?

bindings/SwiftCore.xml
```diff
@@ -211,5 +211,6 @@
         <operator name="&amp;=" operatorKind="Infix" precedenceGroup="AssignmentPrecedence" moduleName="Swift" />
         <operator name="|=" operatorKind="Infix" precedenceGroup="AssignmentPrecedence" moduleName="Swift" />
         <operator name="^=" operatorKind="Infix" precedenceGroup="AssignmentPrecedence" moduleName="Swift" />
+	<typealias name="Foundation.NSRect" type="CoreGraphics.CGRect" />
```
spouliot, line 214: Legit for 64bits :)

/Applications/Xcode_12.5.0-rc.app//Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Foundation.framework/Headers/NSGeometry.h
```
#if __LP64__ || TARGET_OS_IPHONE || NS_BUILD_32_LIKE_64
```

spouliot (approved):

tj-devel709 (approved):

dalexsoto (approved):

rolfbjarne (approved):

SwiftReflector/TypeMapping/TypeDatabase.cs
```diff
@@ -305,6 +316,13 @@ void ReadVersion1_0 (XElement typeDbRoot, ErrorHandling errors)
 				foreach (var op in ops) {
 					AddOperator (op);
 				}
+
+				var aliases = from alias in entityList.Elements ("typealias")
+					      select TypeAliasDeclaration.FromXElement (null, alias);
+
+				foreach (var alias in aliases) {
+					AddTypeAlias (alias);
+				}
```
mandel-macaque, line 325: No need to change, could have been in the LINQ with the ForEach call.

mandel-macaque (approved):

SwiftReflector/CustomSwiftCompiler.cs
stephen-hawley: I don't know if that code can know the current platform. I'll open up an issue to look into it later. It's not so much that the symbol is iOS or OS but that the swift code contains:
```
import Cocoa
```
which AFAIK isn't a real namespace? This particular code is compiler for macOS though.

bindings/SwiftCore.xml
stephen-hawley: Kill me now.

SwiftReflector/TypeMapping/TypeDatabase.cs
stephen-hawley: Or I could have put in an AddRange type call...

PR #667, opened 2021-04-23
Why build and package the compiler when you don't use it?

stephen-hawley:
Remove compiler build step


tj-devel709 (approved):

devops/automation/azure-pipelines.yml
```diff
@@ -116,29 +116,29 @@ jobs:
...
+#      $(Build.SourcesDirectory)/binding-tools-for-swift/devops/automation/build-swift.sh --publish
+#      mkdir -p $(System.DefaultWorkingDirectory)/toolchain
+#      if test -f $(System.DefaultWorkingDirectory)/binding-tools-for-swift/Pack-Man/SwiftToolchain-*.zip; then
+#        cp -c $(System.DefaultWorkingDirectory)/binding-tools-for-swift/Pack-Man/SwiftToolchain-*.zip $(System.DefaultWorkingDirectory)/toolchain
+#        echo "##vso[task.setvariable variable=uploadtoolchain;isOutput=true]1"
+#      else
+#        echo "##vso[task.setvariable variable=uploadtoolchain;isOutput=true]0"
+#      fi
+#    displayName: 'Build Swift'
+#    name: BuildSwift
```
mandel-macaque, line 131: instead of commenting out, you can just disable the step with:
```yaml
enabled: boolean 
```
ref: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/tasks?view=azure-devops&tabs=yaml

Proposed change:
```suggestion
  - script: |
      set -e
      set -x
      $(Build.SourcesDirectory)/binding-tools-for-swift/devops/automation/build-swift.sh --publish
      mkdir -p $(System.DefaultWorkingDirectory)/toolchain
      if test -f $(System.DefaultWorkingDirectory)/binding-tools-for-swift/Pack-Man/SwiftToolchain-*.zip; then
        cp -c $(System.DefaultWorkingDirectory)/binding-tools-for-swift/Pack-Man/SwiftToolchain-*.zip $(System.DefaultWorkingDirectory)/toolchain
        echo "##vso[task.setvariable variable=uploadtoolchain;isOutput=true]1"
      else
        echo "##vso[task.setvariable variable=uploadtoolchain;isOutput=true]0"
      fi
    displayName: 'Build Swift'
    name: BuildSwift
    enabled: false
```

devops/automation/azure-pipelines.yml
```diff
@@ -116,29 +116,29 @@ jobs:
...
-      publishLocation: 'Container'
+#  # This will upload the updated toolchain zip (if any) to Azure Devops
+#  # A later job further down will then publish the toolchain to Azure Blob Storage.
+#  # This is a two-step process because the task that can publish to Azure Blob Storage only works on Windows :/
+#  - task: PublishBuildArtifacts@1
+#    displayName: 'Upload Swift Toolchain'
+#    inputs:
+#      PathtoPublish: $(System.DefaultWorkingDirectory)/toolchain
+#      ArtifactName: 'toolchain'
+#      publishLocation: 'Container'
```
mandel-macaque, line 141: Same:
```suggestion
  # This will upload the updated toolchain zip (if any) to Azure Devops
  # A later job further down will then publish the toolchain to Azure Blob Storage.
  # This is a two-step process because the task that can publish to Azure Blob Storage only works on Windows :/
  - task: PublishBuildArtifacts@1
    displayName: 'Upload Swift Toolchain'
    enabled: false
    inputs:
      PathtoPublish: $(System.DefaultWorkingDirectory)/toolchain
      ArtifactName: 'toolchain'
      publishLocation: 'Container'
```

devops/automation/azure-pipelines.yml
```diff
@@ -189,65 +189,65 @@ jobs:
     continueOnError: true
     condition: and(succeededOrFailed(), eq('${{ parameters.RunningTests }}', true))
```
mandel-macaque, line 191: Two options:

1. We just remove the entire job
2. Use an if statement.

Since it looks like you don't want to remove the code, do the following: 

1. Add a param at the top of the yaml:

```yaml
- name: uploadToolchain
   type: boolean
   default: false
```

Then, just before the job:
```yaml
- ${{ if eq(uploadToolchain, true) }}:
```
Add the job under the if with an increase indentation.

mandel-macaque (changes requested):

dalexsoto (approved):
I have to agree with Manuel, disabling the step is much better than just commenting it out. After Manuel's 👍

mandel-macaque (approved):
ciao old code!

PR #680, opened 2021-05-05
[DylibBinder] Introducing DylibBinder (part 2)

tj-devel709:
Hello team, thanks to the help of @mandel-macaque and @stephen-hawley, 
I have created a much better version of my DylibBinder tool. 

Binding-tools-for-swift does a lot of really great things, but does not handle dylib files.
While it cannot get all the metadata from the dylib, DylibBinder aids greatly in the binding process
by reflecting on the dylib and giving the user an xml file that contains information about the dylib contents
in a format recognizable by the compiler used in BTFS. 
This still requires work of the user, but takes out a lot of the tedious tasks.

This PR also contains some changes to ease the process for when a user wants to pass in this reflected xml file from
DylibBinder into tom-swifty to create a binding!


DylibBinder/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+			var extra = options.optionsSet.Parse (args);
+
+			if (options.PrintHelp) {
+				options.PrintUsage (Console.Out);
+				return;
+			}
+
+			if (string.IsNullOrEmpty (options.DylibPath)) {
+				Console.WriteLine ("The path to the Dylib was not included. Try using --dylibPath.");
+				return;
```
rolfbjarne, line 20: Maybe return a non-zero exit code if there are failures:

```suggestion
				return 1;
```

this also requires changing the function to return `int`, and update the other return statements.

DylibBinder/StringExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace DylibBinder {
+	internal static class StringExtensions {
+		static readonly string [] BuiltInTypes = { "Bool", "Double", "Float", "UInt", "Int" };
+
+		internal static bool IsPublic (this string s)
+		{
+			return !s.Contains ("_");
```
rolfbjarne, line 11: Using a char is slightly faster:

```suggestion
			return !s.Contains ('_');
```

DylibBinder/StringExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+					sb.Insert (match.Index + offset, "Swift.");
+					offset += sizeOfSwiftInsert;
+				}
+			}
+		}
+
+		static bool IsIntMatchUInt (StringBuilder sb, Match match, int offset)
+		{
+			if (match.Index > 0) {
+				if (sb.ToString () [match.Index - 1 + offset] == 'U') {
```
rolfbjarne, line 47: I think you can just index the StringBuilder:

```suggestion
				if (sb [match.Index - 1 + offset] == 'U') {
```

DylibBinder/StringExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+				}
+				return false;
+			}
+			return false;
+		}
+
+		static bool ContainsSwift (StringBuilder sb, Match match, int sizeOfSwiftInsert)
+		{
+			if (match.Index > sizeOfSwiftInsert) {
+				var substring = sb.ToString ().Substring (match.Index - sizeOfSwiftInsert, sizeOfSwiftInsert);
```
rolfbjarne, line 58: You don't seem to use the `substring` variable?

DylibBinder/StringExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+				if (sb.ToString ().Substring (match.Index - sizeOfSwiftInsert, sizeOfSwiftInsert) != "Swift.") {
```
rolfbjarne, line 59: StringBuilder.ToString can return a sub string:

```suggestion
				if (sb.ToString (match.Index - sizeOfSwiftInsert, sizeOfSwiftInsert) != "Swift.") {
```

DylibBinder/XmlGenerator.cs
```diff
@@ -0,0 +1,289 @@
...
+			writer.WriteEndElement ();
+		}
+
+		void WriteGenericParameters (DBGenericParameters GenericParameterCollection)
+		{
+			var validGenericParameters = from g in GenericParameterCollection.GenericParameterCollection
+										 where g.Depth > 0
+										 select g;
+
+			if (validGenericParameters.Count () == 0)
```
rolfbjarne, line 132: `Any` is faster, since it won't have to count all elements, just find one (or none):

```suggestion
			if (!validGenericParameters.Any ())
```

SwiftReflector/NewClassCompiler.cs
```diff
@@ -5291,6 +5300,7 @@ static bool IsUnsafeMutablePointer (SwiftBoundGenericType bgt)
 			return ct.ClassName.ToFullyQualifiedName (true) == "Swift.UnsafeMutablePointer";
 		}
 
+		// TJ - adding isLibrary bool
```
rolfbjarne, line 5303: I think you can remove these comments

rolfbjarne (changes requested):

DylibBinder/Program.cs
tj-devel709: Should I add somewhere that checks this return value? Or is this for the future if anything else calls into my Main?

DylibBinder/StringExtensions.cs
tj-devel709: Hmm when I try to use `return !s.Contains ('_');` I get an error saying that it cannot convert from 'char' to 'string'
and it asks if I want to include Linq and use it's Contains which I don't think is what you are saying correct?

i do see in the docs that string.Contains should contain the Contains (char c) [here](https://docs.microsoft.com/en-us/dotnet/api/system.string.contains?view=net-5.0#System_String_Contains_System_Char_) though so I am not sure what I am missing?

DylibBinder/StringExtensions.cs
tj-devel709: Ah thanks for catching!

DylibBinder/DBGenericParameter.cs
```diff
@@ -0,0 +1,72 @@
...
+		static List<DBGenericParameter> GetGenericParameters (SwiftGenericArgReferenceType refType)
+			=> new () { new DBGenericParameter (refType.Depth, refType.Index) };
+	}
+
+	class DBGenericParameterComparer : EqualityComparer<DBGenericParameter> {
+		public override bool Equals (DBGenericParameter gp1, DBGenericParameter gp2)
+			=> gp1.Name == gp2.Name;
+
+		public override int GetHashCode (DBGenericParameter gp)
+			=> gp.Name.GetHashCode ();
```
spouliot, line 70: same

DylibBinder/DBTypeDeclaration.cs
```diff
@@ -0,0 +1,200 @@
...
+				if (parameter.HasInstance || parameter.IsConstructor) {
+					ApplyInouts (parameter);
+				}
+			}
+		}
+
+		void ApplyInouts (DBParameter parameter)
+		{
+			var sb = new StringBuilder ();
+			sb.Append ($"inout {Module}.{Name}{GenericParametersToString ()}");
```
spouliot, line 87: the nice thing about `StringBuilder` is to be able to reduce temporary string allocations, like what `String.Format` would do. However if you use `$` you end up (very likely) with a `String.Format` call

Instead do

```csharp
sb.Append ("inout ").Append (module).Append ('.').Append (Name).Append (GenericParametersToString ());
```

DylibBinder/DBTypeDeclaration.cs
```diff
@@ -0,0 +1,200 @@
...
+
+		string GenericParametersToString ()
+		{
+			var sb = new StringBuilder ();
+			var sortedGenericParameters = GenericParameters.GenericParameterCollection.OrderBy (gp => gp.Index);
+			foreach (var gp in sortedGenericParameters) {
+				if (sb.Length > 0)
+					sb.Append ($", {gp.Name}");
+				else
+					sb.Append ($"<{gp.Name}");
```
spouliot, line 99: same avoid `.Append ($...)` or `AppendFormat`, like

```csharp
if (sb.Length > 0)
   sb.Append (", ");
else
   sb.Append ('<');
sb.Append (gp.Name);
```

DylibBinder/DBTypeDeclaration.cs
```diff
@@ -0,0 +1,200 @@
...
+			foreach (var protocolContents in protocolContentList) {
+				if (protocolContents.Name.ToFullyQualifiedName ().IsPublic () && !ignoredTypes.Contains (protocolContents.Name.ToFullyQualifiedName ()))
+					TypeDeclarationCollection [module].Add (new DBTypeDeclaration (protocolContents));
+			}
+		}
+
+		SortedSet<string> GetIgnoredTypes ()
+		{
+			var fromTestPath = Path.Combine (Directory.GetCurrentDirectory (), "../../../../DylibBinder/IgnoreList.txt");
+			var fromVSPath = Path.Combine (Directory.GetCurrentDirectory (), "../../IgnoreList.txt");
```
spouliot, line 156: do they differ ?
if not then why not embed them into the assembly ?
you can still accept something else from the command-line

spouliot: > Do you mean to make this path an argument from the command the line?

No :)

You can include files inside an assembly (as resources [1]) and then get a `Stream` back to those at runtime. That way you do not have _find them_ in different locations at runtime. 

That's useful if you don't plan to modify the files without rebuilding them assembly (which the CI does every time). So unless it's meant to be user-edited (which I guess is why you also have a command-line option) then embedding your copy sounds useful.

[1] `csc` arguments to embed a file: https://github.com/xamarin/xamarin-macios/pull/11424/files

DylibBinder/Program.cs
rolfbjarne: This change means that whenever someone uses this executable from scripts or the command line, the executable will return a non-zero exit code (and thus indicate failure), so that scripts can realize that something went wrong.

In other words: there's no need to add checks anywhere.

DylibBinder/DBTypeDeclaration.cs
stephen-hawley: The skip list(s) should be passed in as arguments in the command line.
The main target of this project is going to be `libswiftCore.dylib` (and any other swift-only library without a .swiftmodule file). I don't think it should ever be read in from static strings or an embedded resource.

DylibBinder/IgnoreList.txt
```diff
@@ -0,0 +1,41 @@
...
+Swift.OpaquePointer
+Swift.UnsafeRawBufferPointer
+Swift.UnsafeMutableRawBufferPointer
+Swift.UnsafeRawPointer
+Swift.UnsafeMutableRawPointer
+Swift.CustomStringConvertible
+Swift.AnyObject
+Swift.Any
+Swift.Hashable
+Swift.Equatable"
```
stephen-hawley, line 36: I don't think the `"` is wanted here.

DylibBinder/XmlGenerator.cs
```diff
@@ -0,0 +1,289 @@
...
+using System.Linq;
+using System.Xml;
+using SwiftReflector.SwiftXmlReflection;
+
+namespace DylibBinder {
+	sealed internal class XmlGenerator : IDisposable {
+		XmlWriter writer;
+		private bool _disposed = false;
+
+		public XmlGenerator (DBTopLevel dBTopLevel, string outputPath)
```
stephen-hawley, line 12: While this works, I get twitchy when I see all the work being done in the constructor _and_ when the class is `IDisposable`.  In looking at your code, it looks like it doesn't need to be.
You could instead structure the class like this:

```
sealed internal class XmlGenerator {
    string outputPath;
    XmlWriter writer;

    public XmlGenerator (string outputPath) {
        this.outputPath = outputPath; // throw on null kthx
    }

    public void WriteDBToFile (DBTopLevel dBTopLevel) {
        using (writer = CreateWriter (outputPath)) {
            WriteIntro (dBTopLevel);
            WriteModule (dBTopLevel);
            CloseWriter ();
        }
    }
// etc
}
```

When when you use it, you can do:
```
var generator = new XmlGenerator (outputPath);
generator.WriteDBToFile (db);
```
Which makes the flow clear for what you're doing:
I'm making an XmlGenerator, then I use it to write the the database. In addition, the writer, which is a transient resource, gets managed where it gets used rather than shoving it to the client. Another way of looking at it is to think of the implementation of writing as a mystery to the caller (as it should be). I call a method and it just works, I don't care how it gets done. When you make the class disposable, you're pushing the implementation details out to the caller who now needs to care about an implementation detail.

Now, this is just one way of looking at this from an API point of view. There are several other ways that you can do this.

For example, maybe the class holds onto an instance of the DB and you pass in the path to the method that does the work.

Another way to do this instead is to make the constructor (above) private and public a static method:

```
public static void WriteDBToFile (DBTopLevel db, string outputPath) {
    var generator = new XmlGenerator (outputPath);
    generator.WriteDBToFile (db);
}
```
The the callers do `XmlGenerator.WriteDBToFile (db, outputPath);`

stephen-hawley (changes requested):

DylibBinder/IgnoreList.txt
tj-devel709: You are correct! Thank you!

DylibBinder/StringExtensions.cs
tj-devel709: image
CS1503: Argument 1: Cannot convert from 'char' to 'string'

DylibBinder/DBTypeDeclaration.cs
tj-devel709: Okay I added the skip list to be an input argument!

DylibBinder/XmlGenerator.cs
tj-devel709: Okay I made those changes, thank you!

DylibBinder/StringExtensions.cs
dalexsoto: @tj-devel709 is right, the char overload is not available in .NET Framework 4.8 https://docs.microsoft.com/en-us/dotnet/api/system.string.contains?view=netframework-4.8

rolfbjarne: Ah, that makes sense.

The old way of doing this was to call `string.IndexOf (char) >= 0`(https://docs.microsoft.com/en-us/dotnet/api/system.string.indexof?view=netframework-4.8#System_String_IndexOf_System_Char_)

No need to change this, the current code is more legible.

rolfbjarne (approved):

dalexsoto (approved):

DylibBinder/CheckInventory.cs
```diff
@@ -0,0 +1,55 @@
...
+		public SortedSet<ClassContents> Enums { get; } = SortedSetExtensions.Create<ClassContents> ();
+		public SortedSet<ProtocolContents> Protocols { get; } = SortedSetExtensions.Create<ProtocolContents> ();
+	}
+
+	internal class CheckInventoryDictionary {
+		public SortedDictionary<string, CheckInventoryValues> CheckInventoryDict { get; } = new SortedDictionary<string, CheckInventoryValues> ();
+
+		public CheckInventoryDictionary (ModuleInventory mi)
+		{
+			GetValues (mi);
```
mandel-macaque, line 19: What if mi == null?

DylibBinder/DBAssociatedType.cs
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	internal class DBAssociatedTypes {
+		public HashSet<DBAssociatedType> AssociatedTypeCollection { get; } = new HashSet<DBAssociatedType> (new DBAssociatedTypeComparer ());
+
+		public DBAssociatedTypes ()
+		{
+		}
+
+		public DBAssociatedTypes (DBTypeDeclaration typeDeclaration)
```
mandel-macaque, line 25: What if typeDeclaration == null

DylibBinder/DBAssociatedType.cs
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	// interface for having an AssociatedTypes Property
+	internal interface IAssociatedTypes {
+		public DBAssociatedTypes AssociatedTypes { get; }
+	}
+
+	internal static class DBAssociatedTypesExtensions {
+		public static List<DBAssociatedType> GetAssociatedTypes (this SwiftBaseFunctionType signature)
+			=> signature.EachParameter.SelectMany (t => AssociatedTypesSwitch (t)).ToList ();
```
mandel-macaque, line 38: I think we can do INeumerables in all this methods, create an issue to do that and get a little better. No need for this PR, please reply with the issue to this comment.

DylibBinder/DBAssociatedType.cs
```diff
@@ -0,0 +1,72 @@
...
+		{
+			if (at1 == null && at2 == null)
+				return true;
+			else if (at1 == null || at2 == null)
+				return false;
+			return at1.Name == at2.Name;
+		}
+
+		public override int GetHashCode (DBAssociatedType at)
+			=> at == null ? "null".GetHashCode () : at.Name.GetHashCode ();
```
mandel-macaque, line 70: If you get null, there is not need to do that. GetHashCode needs to return always the same value for the same object, that is all. You could have done:

```
if (at == null)
   return 42;
return at.Name.GetHashCode ();
```
GetHashCode is not guaranteed to return unique hash codes as per MSD docs:

> Two objects that are equal return hash codes that are equal. However, the reverse is not true: equal hash codes do not imply object equality, because different (unequal) objects can have identical hash codes. Furthermore, .NET does not guarantee the default implementation of the GetHashCode method, and the value this method returns may differ between .NET implementations, such as different versions of .NET Framework and .NET Core, and platforms, such as 32-bit and 64-bit platforms. For these reasons, do not use the default implementation of this method as a unique object identifier for hashing purposes. Two consequences follow from this:
> You should not assume that equal hash codes imply object equality.
> You should never persist or use a hash code outside the application domain in which it was created, because the same object may hash across application domains, processes, and platforms.

That is from: https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-5.0#remarks

DylibBinder/DBFunc.cs
```diff
@@ -0,0 +1,106 @@
...
+
+		public DBParameterLists ParameterLists { get; }
+		public DBGenericParameters GenericParameters { get; }
+		public DBAssociatedTypes AssociatedTypes { get; } = new DBAssociatedTypes ();
+
+		public DBFunc (TLFunction tlf, bool isMutating = false, bool isProtocol = false)
+		{
+			Name = tlf.Name.Name;
+			IsStatic = GetStaticStatus (tlf.Signature);
+			IsProperty = false;
```
mandel-macaque, line 38: Looking at this, IsProperty, is it ever true for this type?

DylibBinder/DBParameter.cs
```diff
@@ -0,0 +1,111 @@
...
+			HasInstance = parameterOptions.HasFlag (ParameterOptions.HasInstance);
+			IsConstructor = parameterOptions.HasFlag (ParameterOptions.IsConstructor);
+			IsEmptyParameter = parameterOptions.HasFlag (ParameterOptions.IsEmptyParameter);
+		}
+
+		public static DBParameter CreateInstanceParameter () => new DBParameter ("", "", "self", 0, ParameterOptions.HasInstance);
+		public static DBParameter CreateInstanceParameter (int index) => new DBParameter ("", "", "self", index, ParameterOptions.HasInstance);
+		public static DBParameter CreateConstructorParameter (int index) => new DBParameter (".Type", "", "self", index, ParameterOptions.IsConstructor);
+		public static DBParameter CreateGetterParameter () => new DBParameter (null, null, null, 0, ParameterOptions.IsEmptyParameter);
+		public static DBParameter CreateSetterParameter () => new DBParameter ("", "", "self", 0, ParameterOptions.HasInstance);
```
mandel-macaque, line 32: Lets use String.Empty instead of "" in all these methods. It will be nicer.

DylibBinder/DBParameter.cs
```diff
@@ -0,0 +1,111 @@
...
+				return;
+			}
+
+			foreach (var parameter in signature.EachParameter) {
+				var type = SwiftTypeToString.MapSwiftTypeToString (parameter);
+				var publicName = "_";
+				var privateName = "privateName";
+
+				if (parameter.Name != null)
+					publicName = privateName = parameter.Name.Name;
```
mandel-macaque, line 58: Same can be achieved with ??

DylibBinder/DBParameter.cs
```diff
@@ -0,0 +1,111 @@
...
+			if (hasInstance) {
+				ParameterListCollection.Add (new DBParameterList (signature, true, false, parameterListIndex));
+				parameterListIndex++;
+			} else if (signature.IsConstructor) {
+				ParameterListCollection.Add (new DBParameterList (signature, false, true, parameterListIndex));
+				parameterListIndex++;
+			}
+
+			ParameterListCollection.Add (new DBParameterList (signature, false, false, parameterListIndex));
+			AssociatedTypes.AssociatedTypeCollection.UnionWith (ParameterListCollection.GetChildrenAssociatedTypes ());
```
mandel-macaque, line 98: Is this suppose to execute always or as an 'else' for the other two branches of the if?

DylibBinder/DBParameter.cs
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		public DBParameterLists (bool hasInstance, string propertyType)
+		{
+			var parameterListIndex = 0;
+			if (hasInstance) {
+				ParameterListCollection.Add (new DBParameterList (propertyType, true, parameterListIndex));
+				parameterListIndex++;
+			}
+			ParameterListCollection.Add (new DBParameterList (propertyType, false, parameterListIndex));
```
mandel-macaque, line 108: Same question.

DylibBinder/DBProperty.cs
```diff
@@ -0,0 +1,57 @@
...
+		{
+			Name = propertyContents.Name.Name;
+			IsStatic = propertyContents.Getter.IsStatic;
+			AssociatedTypes.AssociatedTypeCollection.UnionWith (propertyContents.Getter.ReturnType.GetAssociatedTypes ());
+			Type = SwiftTypeToString.MapSwiftTypeToString (propertyContents.Getter.ReturnType, propertyContents.Class.ClassName.Module.Name);
+			GenericParameters = new DBGenericParameters (propertyContents.Getter.ReturnType);
+
+			Getter = new DBFunc (this, "Getter");
+			if (propertyContents.Setter != null)
+				Setter = new DBFunc (this, "Setter");
```
mandel-macaque, line 33: Try to see if with ?? you can write nicer code (tip: you can)

mandel-macaque (changes requested):
Small changes and a few questions. Good work.

DylibBinder/DBAssociatedType.cs
tj-devel709: https://github.com/xamarin/binding-tools-for-swift/issues/686

DylibBinder/DBAssociatedType.cs
tj-devel709: Makes sense! Thanks!

DylibBinder/DBFunc.cs
tj-devel709: For this constructor, it is false, but for the constructor that takes a DBProperty, it is true

DylibBinder/DBParameter.cs
tj-devel709: These should always happen!

DylibBinder/DBProperty.cs
tj-devel709: So I read up on the null coalescing operator and I'm having trouble applying it here.
I want to check the if propertyContents.Setter is null but I don't want to set as the value of Setter if it is not null.

DylibBinder/DBParameter.cs
tj-devel709: I am also confused on how to use ?? here.
I want to check parameter.Name but I don't want to set that value. Instead I want to set the value to be parameter.Name.Name

mandel-macaque (approved):

stephen-hawley (approved):

PR #681, opened 2021-05-05
Make sandwiches run

stephen-hawley:
This sample wasn't running. To fix this, I turned the sample code into a unit test then when it ran correctly, I played the game "one of these things is not like the other" and turned the main rule and the `runit` rule into what was happening in the unit test.
I spent some time stripping out bits that did nothing and tries to optimize it for readability since this is one of the most likely bits of the samples to actually get read.

Will migrate these to the other failing samples in the next PR.

### DIFF
--- a/samples/sandwiches/Makefile
+++ b/samples/sandwiches/Makefile
@@ -5,7 +5,7 @@ SWIFT_LIB = $(XCODE_DEV)/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-5.0
 SWIFTC = $(SWIFT_BIN)/swiftc
 SDK = `xcrun --sdk macosx --show-sdk-path`
 SWIFTARGS = -emit-library -enable-library-evolution -emit-module -emit-module-interface \
-	-sdk $(XCODE_DEV)/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk \
+	-sdk $(XCODE_DEV)/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk \
 	-L $(SWIFT_LIB)/macosx \
 	-framework XamGlue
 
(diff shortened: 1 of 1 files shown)


samples/sandwiches/Makefile
```diff
@@ -1,9 +1,13 @@
 SWIFT_BIN = /usr/bin
-SWIFT_LIB = /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-5.0
+XCODE_DEV = /Applications/Xcode.app/Contents/Developer
+SWIFT_LIB = $(XCODE_DEV)/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-5.0
 
 SWIFTC = $(SWIFT_BIN)/swiftc
 SDK = `xcrun --sdk macosx --show-sdk-path`
-SWIFTARGS = -v -sdk $(SDK) -L $(SWIFT_LIB) -emit-module -emit-library -enable-library-evolution -emit-module-interface
+SWIFTARGS = -emit-library -enable-library-evolution -emit-module -emit-module-interface \
+	-sdk $(XCODE_DEV)/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk \
```
spouliot, line 8: you might have to 
* move `11.1` to a variable right after the `XCODE_*` ones; or
* use `MacOSX.sdk` which should be a symlink to `MacOSX11.1.sdk` for your Xcode (and 11.3.sdk for mine)

I prefer the latter but I recall Apple broke it in the past... still likely to require less updates over time

spouliot (approved):

rolfbjarne (approved):
👍 once Sebastien's concern is fixed.

samples/sandwiches/Makefile
stephen-hawley: Good idea!

PR #696, opened 2021-06-04
First cut at framework making tool

stephen-hawley:
This is a first stab at a shell script to generate .frameworks and .xcframeworks. Please read everything here before commenting. I'm going to break this down into sections to help reviewers understand everything that's going on in here.

### What will this tool do?
This tool compiles a set of `.c` files and a set of `.swift` files into for a set of CPU architectures targeting a particular operating system for either device or simulator or both. It them combines the output into a single `.framework` (and in the future) it will combine these into a single `.xcframework`.

### Why is this tool needed?
This tool is needed because the addition of Apple silicon means that for iOS, a complete fat library is no longer viable since it could contain arm64 targeting iOS and arm64 targeting iOS-simulator. The solution is to use an .xcframework and we do not have such a tool.

### Where will this tool be used?
It will be used in the Makefile for XamGlue, which will greatly simplify that Makefile.
It will also be used in BTfS which does a lot of this work manually. In addition to consuming `.framework` files, BTfS should consume `.xcframework` files and generate the wrappers in a `.xcframework` files as well.
In addition, this tool does much of the work in parallel.
In the future, I anticipate having to build versions of XamGlue which target multiple minimum operating system versions.
That will become significantly easier with this tool.

### How does this tool work?
It works in a series of steps:

1. Make a `.framework` for simulator
2. Make a `.framework` for devices
3. Merge the `.framework`s into a `.xcframework` if requested (TODO)

To make a `.framework` the code does the following for every architecture:
1. Compile the `.c` files into `.o` files
2. Compile the `.swift` files and link the `.o` files into a module
3. Build the Info.plist for the module (TODO) 
4. Combine all the modules into a `.framework`.

Note that steps 1 and 2 are treated atomically and get run in parallel across architectures.

A `.framework` has the following structure to it for a given `MODULE`:
```
MODULE.framework/
    Info.plist
    MODULE - the fat binary
    Modules/
    	MODULE.swiftmodule/
    		ARCH.swiftdoc
    		ARCH.swiftinterface
    		ARCH.swiftmodule
    		ARCH.swiftsourceinfo
```
For every ARCH, there will be a set of `.swiftdoc` `.swiftinterface` `.swiftmodule` and `.swiftsourceinfo` files.

### Why the .c file support?
Because XamGlue needs them, plain and simple.

### What's left to do?
1. `.framework` files need to be combined into a single `.xcframework` if requested or the single `.framework` needs to be left at the top level of the build directory.
2. Need to generate Info.plist for each framework.
3. Need to clean up files.
4. Need to handle errors properly.
5. Consider a set of possible install_name_tool steps. I think we can actually eliminate some or all of the ones we do as I *think* they're not needed anymore.

### If there's so much to do, why are you putting in a PR right now?
1. Because this was getting substantial and I care about your sanity
2. Because this I haven't written a bash script of this complexity in a while and I'm sure I'm messing things up
3. It's easier to change direction if I have to now rather than later.

## What are the command line arguments?
```
make-framework --frameworks fw fw fw 
				--libraries fw fw fw fw 
			*	--swift-files sf sf sf sf
				--c-files cf cf cf cf 
				--target-os osname 
			*	--minimum-os-version vers 
				--simulator-archs ar ar ar 
				--device-archs ar ar ar
			*	--module-name name 
				--extra-c-args arg arg arg
				--extra-swift-args arg arg arg
			*	--output-path path
				--make-xcframework
```
The ones marked with a `*` are required. Either one or both of `--simulator-archs` and `--device-archs` are required.

### DIFF
--- a/tools/make-framework
+++ b/tools/make-framework
@@ -1,0 +1,418 @@
+#!/bin/bash
+
+COLOR_RED=$(tput setaf 1 2>/dev/null || true)
+COLOR_ORANGE=$(tput setaf 3 2>/dev/null || true)
+COLOR_MAGENTA=$(tput setaf 5 2>/dev/null || true)
+COLOR_BLUE=$(tput setaf 6 2>/dev/null || true)
+COLOR_CLEAR=$(tput sgr0 2>/dev/null || true)
+COLOR_RESET=uniquesearchablestring
+
+usage () {
+	echo "Compiles a set of swift files into a framework and optionally"
+	echo "groups all the frameworks into an xcframework."
+	echo
+	echo "arguments:"
+	echo "  --frameworks fw1 fw2 ..."
+	echo "    adds framework directories to swift compilation (-F) (optional)"
+	echo " --libraries lb1 lb2 ..."
+	echo "    adds library directories to swift compilation (-L) (optional)"
+	echo " --swift-files sf1 sf2 ..."
+	echo "    adds swift files to be compiled (required)"
+	echo " --c-files cf1 cf2 ..."
+	echo "    adds c files to be compiled (optional)"
+	echo " --target-os os-name, one of ios, tvos, watchos, macosx (required)"
+	echo "    sets the target operating system for the build."
+	echo " --minimum-os-version version (required)"
+	echo "    sets the minimum operating system version for the compilation."
+	echo " --simulator-archs arch1 arch2 ..."
+	echo "    sets the architectures for a simulator build."
+	echo " --device-archs arch1 arch2 ..."
+	echo "    sets the architectures for a device build."
+	echo " --module-name name (required)"
+	echo "    sets the name of the output module."
+	echo " --extra-c-args arg1 arg2 ..."
+	echo "    sets extra arguments to pass to the c compiler."
+	echo " --extra-swift-args arg1 arg2 ..."
+	echo "    sets extra arguments to pass to the swift compiler."
+	echo " --output-path path (required)"
+	echo "    sets the directory where there final output will live."
+	echo " --make-xcframework (optional)"
+	echo "    if present, puts both device and simulator builds into an xcframework"
+	echo "    if present, both --simulator-archs and --device-archs must be present."
+	echo " --verbose"
+	echo "    if present, be more talky."
+	echi "    if present, more than once be super talky."
+	echo " --help"
+	echo "    prints this message."
+        echo "This tool compiles swift files into a framework with the given"
+	echo "module name. Optionally it will combine the frameworks into an"
+	echo ".xcframework directory."
+}
+
+usageexit () {
+	usage
+	exit 1
+}
+
+
+while (( "$#" )); do
+	case "$1" in
+	--frameworks)
+		shift
+		while (( "$#" )); do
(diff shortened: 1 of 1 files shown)


tools/make-framework
```diff
@@ -0,0 +1,418 @@
...
+	echo " --extra-swift-args arg1 arg2 ..."
+	echo "    sets extra arguments to pass to the swift compiler."
+	echo " --output-path path (required)"
+	echo "    sets the directory where there final output will live."
+	echo " --make-xcframework (optional)"
+	echo "    if present, puts both device and simulator builds into an xcframework"
+	echo "    if present, both --simulator-archs and --device-archs must be present."
+	echo " --verbose"
+	echo "    if present, be more talky."
+	echi "    if present, more than once be super talky."
```
dalexsoto, line 44:
```suggestion
	echo "    if present, more than once be super talky."
```

tools/make-framework
```diff
@@ -0,0 +1,418 @@
+#!/bin/bash
```
rolfbjarne, line 1: I suggest using `-e`, that way errors aren't silently ignored:

```shell
#!/bin/bash -e
```

tools/make-framework
```diff
@@ -0,0 +1,418 @@
...
+	fi
+	for arch in "$@"
+	do
+		copy_arch_files $arch "$builddir" "$swiftmoduledir"
+		sourcemodules="$sourcemodules ${builddir}/$arch/$module_name"
+	done
+	if [[ -n $superverbose ]]; then
+		echo lipo -create $sourcemodules -output "${frameworkdir}/${module_name}"
+	fi
+}
```
rolfbjarne, line 386: You say you call lipo ("echo lipo...") but it doesn't look like you ever do?

rolfbjarne (changes requested):
Do you know about `shellcheck`? I've found that it's very helpful in writing better bash scripts (and getting them less wrong the first time): https://gist.github.com/rolfbjarne/ef4440a5d8a87b09a589013631689720

tools/make-framework
stephen-hawley: 👍  That's in my list of TODOs.
Will look at shellcheck.

stephen-hawley: I put in most of the spellcheck changes. Most of them were quotations

stephen-hawley: This is about as far as this can go without addressing other issues like how it's packaged and where it will live.

rolfbjarne: Not sure I follow: do you mean to say that you can't add `-e` yet, because you'd run into errors you want to ignore?

stephen-hawley: For that, I haven't decided whether I want `-e` and fail hard or if I want to manually handle the errors. This script will be run from both the build and from the internals of BTfS.

rolfbjarne: You can always decide on a case-to-case basis.

One way to ignore failures is to add `|| true` to commands you don't care about:

```shell
some command that may fail || true
```

Or you can print it out:

```shell
if ! some command that may fail; then
    echo "Some command that may fail failed, but we don't care."
fi
```

rolfbjarne (approved):

Issue #702, opened 2021-06-05
[DylibBinder] Generic Parameter Issue


tj-devel709:
In DynamicXmlTests - TestGenericMethodInGenericClass the Generic parameters get a little wonky.

**DylibBinder's xml**
```
<?xml version="1.0" encoding="utf-8"?>
<xamreflect version="1.0">
   <modulelist>
      <module name="SomeModule" swiftVersion="5.0">
         <typedeclaration kind="class" name="Foo" accessibility="Public" isDeprecated="false" isUnavailable="false" isObjC="false" isFinal="false">
            <members>
               <func name=".ctor" returnType="SomeModule.Foo&lt;T0&gt;" isProperty="false" hasThrows="false" isStatic="false" isMutating="false" operatorKind="None" accessibility="Public" isDeprecated="false" isUnavailable="false" isFinal="false" isOptional="false" isRequired="false" isConvenienceInit="false" objcSelector="" isPossiblyIncomplete="false">
                  <parameterlists>
                     <parameterlist index="0">
                        <parameter index="0" type="inout SomeModule.Foo&lt;T0&gt;.Type" publicName="" privateName="self" isVariadic="false" />
                     </parameterlist>
                     <parameterlist index="1">
                        <parameter index="0" type="T0" publicName="a" privateName="a" isVariadic="false" />
                     </parameterlist>
                  </parameterlists>
               </func>
               <func name=".nctor" returnType="SomeModule.Foo&lt;T0&gt;" isProperty="false" hasThrows="false" isStatic="false" isMutating="false" operatorKind="None" accessibility="Public" isDeprecated="false" isUnavailable="false" isFinal="false" isOptional="false" isRequired="false" isConvenienceInit="false" objcSelector="" isPossiblyIncomplete="false">
                  <parameterlists>
                     <parameterlist index="0">
                        <parameter index="0" type="inout SomeModule.Foo&lt;T0&gt;.Type" publicName="" privateName="self" isVariadic="false" />
                     </parameterlist>
                     <parameterlist index="1">
                        <parameter index="0" type="T0" publicName="a" privateName="a" isVariadic="false" />
                     </parameterlist>
                  </parameterlists>
               </func>
[...]

Issue #701, opened 2021-06-05
[DylibBinder] Convenience vs Designated Constructor


tj-devel709:
Need to look over the difference between the two and how to handle them
Fails DynamicXmlTests - TestConvenienceCtor

Issue #700, opened 2021-06-05
[DylibBinder] Destructors


tj-devel709:
Make sure to add in destructors
DynamicXmlTests - TestClassHasDestructor

Issue #699, opened 2021-06-05
[DylibBinder] Constructor Behavior


tj-devel709:
I guess not all constructors are meant to be inout?
Example) DynamicXmlTests TestCtorType

PR #703, opened 2021-06-05
[DylibBinder] Adding tests

tj-devel709:
This PR runs the DynamicXmlTests using the DylibBinder for reflection!

- Passes 34 Tests
- Fails 57 Tests 

A handful of tests fail for small things like order of items or the public name being incorrect.
I also redid some things in DylibBinder to fix major issues such as global functions being accounted for.
Each test that fails is linked to an issue.

Issue containing large list of things we cannot do in DylibBinder yet: https://github.com/xamarin/binding-tools-for-swift/issues/697
Generic Parameter Issue: https://github.com/xamarin/binding-tools-for-swift/issues/702
Constructor Behaviour Issue: https://github.com/xamarin/binding-tools-for-swift/issues/699
Protocols not fully supported Issue: https://github.com/xamarin/binding-tools-for-swift/issues/698
Destructor Issue: https://github.com/xamarin/binding-tools-for-swift/issues/700
Convenience vs Designated Constructor Issue: https://github.com/xamarin/binding-tools-for-swift/issues/701


DylibBinder/CheckInventory.cs
```diff
@@ -56,5 +54,19 @@ void GetProtocols (ModuleInventory mi, SwiftName module)
...
 		}
+
+		public static List<OverloadInventory> GetGlobalFunctions (ModuleInventory mi, SwiftName module)
+		{
+			Exceptions.ThrowOnNull (mi, nameof (mi));
+			Exceptions.ThrowOnNull (module, nameof (module));
+			var functions = new List<OverloadInventory> ();
+
+			foreach (var f in mi.FunctionsForName (module)) {
+				if (!f.Name.Name.Contains ("_")) {
```
stephen-hawley, line 65: what will happen with:
```
public func this_is_cool() -> Int {
    return 42
}
```
?

DylibBinder/DBFunc.cs
```diff
@@ -32,7 +32,7 @@ internal class DBFunc : IAssociatedTypes {
 		public DBGenericParameters GenericParameters { get; }
 		public DBAssociatedTypes AssociatedTypes { get; } = new DBAssociatedTypes ();
 
-		public DBFunc (TLFunction tlf, bool isMutating = false, bool isProtocol = false)
+		public DBFunc (TLFunction tlf, bool isMutating = false, bool isProtocol = false, bool isGlobal = false)
```
stephen-hawley, line 35: This function shouldn't need an `isGlobal` flag. You should get the same result from doing `tlf.IsTopLevelFunction`.
The `TLFunction` type is a holder for functions defined in a dylib. Basically, the `TL` stands for top level, but what happens when you get a member function? Then the member is defined on a class and the `TLFunction` will have a class associated with it and with a nesting name length > 0. This predicate checks that for you.

stephen-hawley (changes requested):

DylibBinder/CheckInventory.cs
tj-devel709: I realized that I made an extension method in StringExtensions called `IsPublic (this string s)` that I use elsewhere in this file for the same purpose. I made sure all 3 times, I use this method. Then I linked this issue in that method.

In the issue I included the private and internal classes, enums, structs, protocols, and functions that have the underscore. I need to find another way to fix that!

stephen-hawley: This should be doable without needing to look at the name. I'll talk to you about this directly.

PR #706, opened 2021-06-11
Added code to make Info.plist, make the framework, and clean up.

stephen-hawley:
Updated the script to do the following:
1 - find plist-swifty.exe relative to the script
2 - call it to generate the Info.plist file
3 - call xcodebuild to make the xcframework itself OR copy the output framework to the build directory
4 - clean up

### DIFF
--- a/tools/make-framework
+++ b/tools/make-framework
@@ -50,6 +50,11 @@ usage () {
         echo "This tool compiles swift files into a framework with the given"
 	echo "module name. Optionally it will combine the frameworks into an"
 	echo ".xcframework directory."
+	echo "\nIf you set the envonment variable PLIST_SWIFTY to the path to"
+	echo "the tool plist-swift.exe, make-framework will use it. If the variable"
+	echo "is not set, make-framework will look for the tool relative to"
+	echo "this script. If the --info-plist-file argument is set, then"
+	echo "PLIST_SWIFTY is ignored."
 }
 
 usageexit () {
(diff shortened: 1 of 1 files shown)


tools/make-framework
```diff
@@ -243,6 +248,21 @@ if [[ -n "$sim_and_device" && -z "$make_xcframework" ]]; then
 	exit 1
 fi
 
+# this one-liner is from https://stackoverflow.com/questions/59895/how-can-i-get-the-source-directory-of-a-bash-script-from-within-the-script-itsel
```
rolfbjarne, line 251: Totally unimportant, but:

```suggestion
# this one-liner is from https://stackoverflow.com/questions/59895/how-can-i-get-the-source-directory-of-a-bash-script-from-within-the-script-itself
```

tools/make-framework
```diff
@@ -243,6 +248,21 @@ if [[ -n "$sim_and_device" && -z "$make_xcframework" ]]; then
...
+# it has the caveat that it doesn't work on sylinks
```
rolfbjarne, line 252:
```suggestion
# it has the caveat that it doesn't work on symlinks
```

rolfbjarne (approved):

tools/make-framework
stephen-hawley: Oddly enough, this is correct. I checked the link twice because I didn't believe it.

PR #707, opened 2021-06-14
Added representation for a compilation target

stephen-hawley:
This is a simple representation of a compilation target.
These will be built from xcframeworks and frameworks, eventually.

Am I missing a CPU that we care about?

### DIFF
--- a/SwiftReflector/CompilationTarget.cs
+++ b/SwiftReflector/CompilationTarget.cs
@@ -1,0 +1,91 @@
+﻿using System;
+namespace SwiftReflector {
+	public class CompilationTarget {
+		public CompilationTarget (PlatformName os, TargetCpu cpu, TargetEnvironment environment,
+			Version minimumOSVersion, TargetManufacturer manufacturer = TargetManufacturer.Apple)
+		{
+			if (os == PlatformName.None)
+				throw new ArgumentOutOfRangeException (nameof (os));
+			OperatingSystem = os;
+			Cpu = cpu;
+			Environment = environment;
+			if (minimumOSVersion == null)
+				throw new ArgumentNullException (nameof (minimumOSVersion));
+			MinimumOSVersion = minimumOSVersion;
+			Manufacturer = manufacturer;
+		}
+
+		public PlatformName OperatingSystem {
+			get; private set;
+		}
+
+		public TargetCpu Cpu {
+			get; private set;
+		}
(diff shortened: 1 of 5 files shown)


SwiftReflector/Enums.cs
```diff
@@ -149,5 +149,22 @@ public enum ReflectionStrategy {
...
+		X86_64,
+	}
+
+	public enum TargetManufacturer {
+		Apple,
+	}
+
+	public enum TargetEnvironment {
+		Device,
+		Simulator,
```
spouliot, line 167: It's often called _variant_
and it's also used for Catalyst, being an macOS variant

Right now I'm not quite sure how you would represent this...

tj-devel709 (approved):

rolfbjarne (approved):

SwiftReflector/CompilationTarget.cs
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		string CpuToString ()
+		{
+			switch (Cpu) {
+			case TargetCpu.Arm64: return "arm64";
+			case TargetCpu.Armv7: return "armv7";
+			case TargetCpu.Armv7s: return "armv7s";
+			case TargetCpu.I386: return "i386";
+			case TargetCpu.X86_64: return "x86_64";
```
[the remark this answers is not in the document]
stephen-hawley: Thanks!

SwiftReflector/Enums.cs
stephen-hawley: What's the extended target triple for that?

spouliot: I don't recall... that was not my area of interest back then :)

A quick google suggest `x86_64-apple-ios-macabi` has being used (and that rings a bell, similar to what I used for xtro). Replace first part with `arm64` when needed.

The xcframework store the binaries inside a `ios-arm64_x86_64-maccatalyst` directory, see https://github.com/spouliot/xcframework

spouliot (approved):

SwiftReflector/Enums.cs
stephen-hawley: I think it will just come down to adding a `MacCatalyst` to the TargetEnvironment. It will require a change to the build script, but LA LA LA I CAN'T HEAR YOU

Issue #708, opened 2021-06-17
MachO might need to be up to date


stephen-hawley:
It doesn't detect Arm64_32 and it probably should.
Update MachO helpers to handle that.
While you're there update main code to handle Arm7 and Arm64e.

PR #710, opened 2021-06-22
Let's actually compile some stuff

stephen-hawley:
Here's what we get here -

There is a class called CompilationSettings which represents extra the bits and pieces we're going to need to compiling swift code.

Unlike previously, instead of building up all the various settings that are needed to match a given existing library, the settings includes the `UniformTargetRepresentation` of a (possibly) existing library, framework, or xcframework. Then when you supply the swift files, libraries, frameworks, and module name the `CompilationSettings` will conjure up the arguments needed to correctly run `make-framework` in such a way to match the `UniformTargetRepresentation`.

In addition, I made `MachOHawley.cs` accept arm64e, and arm64_32 fixing issue [708](https://github.com/xamarin/binding-tools-for-swift/issues/708).

Unit tests verify that this code builds a library for macOS, frameworks for iOS and iOS-simulator, and an xcframework for iOS and iOS simulator.

### DIFF
--- a/SwiftReflector/CompilationSettings.cs
+++ b/SwiftReflector/CompilationSettings.cs
@@ -1,0 +1,224 @@
+﻿using System;
+using System.IO;
+using System.Text;
+using System.Collections.Generic;
+using System.Linq;
+using SwiftReflector.IOUtils;
+using SwiftRuntimeLibrary;
+using System.Reflection;
+
+namespace SwiftReflector {
+	public class CompilationSettings {
+		public CompilationSettings (string outputDirectory, string moduleName, UniformTargetRepresentation targetRepresentation,
+			IEnumerable<string> frameworkPaths = null, IEnumerable<string> libraryPaths = null,
+			IEnumerable<string> swiftFilePaths = null, string makeFrameworkScript = null,
+			string workingDirectory = null)
+		{
+			OutputDirectory = Exceptions.ThrowOnNull (outputDirectory, nameof (outputDirectory));
+			ModuleName = Exceptions.ThrowOnNull (moduleName, nameof (moduleName));
+
+			FrameworkPaths = new List<string> ();
+			if (frameworkPaths != null)
+				FrameworkPaths.AddRange (frameworkPaths);
+
+			LibraryPaths = new List<string> ();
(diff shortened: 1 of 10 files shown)


SwiftReflector/CompilationSettings.cs
```diff
@@ -0,0 +1,224 @@
...
+
+			if (SwiftFilePaths.Count == 0)
+				throw new Exception ("No files provided to compile.");
+			CheckAllFilesExist ();
+			var args = new StringBuilder ();
+
+			// useful for debugging:
+			// args.Append (" --verbose --verbose ");
+
+			args.Append ("--output-path ").Append (OutputDirectory);
```
rolfbjarne, line 103: This won't work if `OutputDirectory` contains a space.

I've found that it's usually much better to keep arguments as an array/list until the very end, and only then convert to a single string:

https://github.com/xamarin/xamarin-macios/blob/d11852e02c3d51be71b6989e0eacb07e945b0e82/tools/common/Execution.cs#L68

rolfbjarne (changes requested):

SwiftReflector/CompilationSettings.cs
stephen-hawley: Agree, but in this case 2 changes fixes them all instead of more than a dozen changes.

SwiftReflector/CompilationSettings.cs
```diff
@@ -0,0 +1,225 @@
...
+		{
+			foreach (var file in SwiftFilePaths) {
+				if (!File.Exists (file))
+					throw new FileNotFoundException ("Couldn't find swift source file", file);
+			}
+		}
+
+		string GetScriptPath ()
+		{
+			var path = MakeFrameworkScript ?? Environment.GetEnvironmentVariable ("") ??
```
rolfbjarne, line 182: This looks like a typo?

> Environment.GetEnvironmentVariable ("")

rolfbjarne (approved):

SwiftReflector/CompilationSettings.cs
stephen-hawley: Nice.

Issue #719, opened 2021-07-16
SwiftModuleFinder is obsolete


stephen-hawley:
Most of its functionality has been better implemented by UniformTargetRepresentation.
Just kill it with fire.

Issue #718, opened 2021-07-16
Refactor UniformTargetRepresentation


stephen-hawley:
Here's what you need to do:

Make an interface with all the common methods and properties and then make everything implement those interfaces.
The top UniformTargetRepresentation can also implement that interface and do it in terms of code like:
```
ITargetRepresentation ActualRep {
    get {
        return ((ITargetRepresentation)Library ?? Framework ?? XCFramework);
    }
}
public T SomeInterfaceMethod ()
{
    return ActualRep.SomeInterfaceMethod ();
}
```

Issue #715, opened 2021-07-06
You broke the samples again


stephen-hawley:
Samples aren't building with the new tools. Look in more detail, thanks.

PR #721, opened 2021-07-16
fix up sample builds

stephen-hawley:
Let's make the samples build, fixing issue [715](https://github.com/xamarin/binding-tools-for-swift/issues/715)
What was happening?
The code was having trouble dealing with `.framework` directories and wasn't building the piglatin sample.
Why was this happening?
The SwiftModuleFinder code is woefully naive. Instead, I'm using the `UniversalFrameworkRepresentation`, which has most of that heavy lifting built into it.
To help things along, I added in a constructor to make a CompilationTarget from a string for some of the places that need that.

There are some problems. The UniformTargetRepresentation has some design errors and [we can make that better](https://github.com/xamarin/binding-tools-for-swift/issues/718). `SwiftModuleFinder` [can probably go away entirely](https://github.com/xamarin/binding-tools-for-swift/issues/719).

Those changes don't belong here. I'll do them later.

Added tests for the code to parse a `CompilationTarget` and all the existing tests pass and the samples build.

### DIFF
--- a/SwiftReflector/CustomSwiftCompiler.cs
+++ b/SwiftReflector/CustomSwiftCompiler.cs
@@ -219,13 +219,13 @@ namespace SwiftReflector {
 
 			var modulesInLibraries = SwiftModuleFinder.FindModuleNames (libraryDirectories, CompilerInfo.Target);
 
-			List<ISwiftModuleLocation> locations = SwiftModuleFinder.GatherAllReferencedModules (modulesInLibraries,
-													     includeDirectories, CompilerInfo.Target);
+			var locations = SwiftModuleFinder.GatherAllReferencedModules (modulesInLibraries,
+													     includeDirectories.ToList (), CompilerInfo.Target);
 			string output = "";
 			try {
-				output = Reflect (locations.Select (loc => loc.DirectoryPath), libraryDirectories, pathName, extraArgs, moduleNames);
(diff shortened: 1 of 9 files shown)


rolfbjarne (approved):

SwiftReflector/CustomSwiftCompiler.cs
```diff
@@ -219,13 +219,13 @@ public Version GetCompilerVersion ()
...
-													     includeDirectories, CompilerInfo.Target);
+			var locations = SwiftModuleFinder.GatherAllReferencedModules (modulesInLibraries,
+													     includeDirectories.ToList (), CompilerInfo.Target);
 			string output = "";
 			try {
-				output = Reflect (locations.Select (loc => loc.DirectoryPath), libraryDirectories, pathName, extraArgs, moduleNames);
+				output = Reflect (locations.Select (loc => loc.ParentPath), libraryDirectories, pathName, extraArgs, moduleNames);
 			} finally {
-				locations.DisposeAll ();
+				//locations.DisposeAll ();
```
tj-devel709, line 228: Do you still want this try/finally block?

tj-devel709 (approved):

SwiftReflector/CustomSwiftCompiler.cs
stephen-hawley: No. No I do not. Thanks!

Issue #739, opened 2021-08-06
NSImageViewSmokeTest4 is failing in Xcode 13


stephen-hawley:
Check to see where that's going wrong.

Issue #738, opened 2021-08-06
NSImageView override on Display is failing


stephen-hawley:
Failing the test NSImageViewSmokeTest1.
Check to see what the deal is in the NSImageView maybe?

PR #744, opened 2021-08-10
Fixed two more failing Swift 5.5 tests

stephen-hawley:
Fixes issues [738](https://github.com/xamarin/binding-tools-for-swift/issues/738) and [739](https://github.com/xamarin/binding-tools-for-swift/issues/739).

What's going on:
In Swift 5.5, if you declare an ObjC class the compiler will add a new attribute to some of the declarations: `@_Concurrency.MainActor`. This is an issue because the published swift grammar doesn't allow periods in names, yet here we are. I opened [this issue](https://bugs.swift.org/browse/SR-15050) with Apple so that they fix the grammar.

I adjusted the parser grammar in two ways. The first was to make a separate rule for nominal declarations which includes `init` and `deinit` and removed that from the `declaration` rule. Second, I adjusted the rule for attributes to allow the identifier to have period separators. I do not know if they allow arbitrary depth, but until they correct the grammar, this is what it is.

Tests pass.


spouliot (approved):
`.`

rolfbjarne (approved):

PR #781, opened 2023-10-23
[Tests] Fix SimplestImportTest and SimpleImportTest for module count

stephen-hawley:
We now get another phantom import module - take that into account.
Make assert messages more bettererer

### DIFF
--- a/tests/tom-swifty-test/XmlReflectionTests/SwiftInterfaceParserTests.cs
+++ b/tests/tom-swifty-test/XmlReflectionTests/SwiftInterfaceParserTests.cs
@@ -12,6 +12,7 @@ using SwiftReflector;
 using SwiftReflector.SwiftInterfaceReflector;
 using SwiftReflector.TypeMapping;
 using tomwiftytest;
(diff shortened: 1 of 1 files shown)


tests/tom-swifty-test/XmlReflectionTests/SwiftInterfaceParserTests.cs
```diff
@@ -67,11 +68,24 @@ public func hello ()
...
+
+		static string AllImportModules (List<string> importModules)
+		{
+			var sb = new StringBuilder ();
+			foreach (var module in importModules) {
+				if (sb.Length > 0)
+					sb.Append (", ");
+				sb.Append (module);
+			}
+			return sb.ToString ();
```
mandel-macaque, line 88: Why not?
```suggestion
			return string.Join (", ", importModules);
```

mandel-macaque (approved):
Minor, we could use the BCL api.

rolfbjarne (approved):

PR #789, opened 2023-11-09
trying a marshaling experiment

stephen-hawley:
This is an experiment to find out if we can use "standard" opaque structs in place of our usual marshaling.
This is in preparation for (hopefully) the runtime providing the ability to directly marshal value types of up to 4 machine words (swift ABI) instead of being limited to 2.

Using opaque structs is more elegant and less heavy handed than what we do in most cases, although it is still heavy handed.

In the process, I added a few goodies to Dynamo to generate the code that I wanted in C# more easily.

All tests pass.

### DIFF
--- a/tests/tom-swifty-test/SwiftReflector/ExpermimentalTests.cs
+++ b/tests/tom-swifty-test/SwiftReflector/ExpermimentalTests.cs
@@ -1,0 +1,255 @@
+﻿using System;
+using System.Runtime.InteropServices;
+using Dynamo.CSLang;
+using NUnit.Framework;
+using SwiftRuntimeLibrary;
+using SwiftRuntimeLibrary.SwiftMarshal;
+using tomwiftytest;
+
+namespace SwiftReflector
+{
+	[TestFixture]
+	public class ExpermimentalTests
+	{
+		[Test]
+		public void SingleIntStruct ()
+		{
+			var swiftCode = @"
+@frozen
+public struct OneInt {
+	public var X:Int
+	public init () { X = 17 }
+}
+";
+
+			// C# aux code:
+			// public struct XXOpaque1 {
+			// nint x0
+			// public static unsafe nint Getter (OneInt self)
+			// {
+			//    fixed (byte* thisSwiftDataPtr = SwiftRuntimeLibrary.SwiftMarshal.StructMarshal.Marshaler.PrepareValueType (self)) {
+			//        var p = (XXOpaque1 *)thisSwiftDataPtr;
+			//        return XGetter (*p);
+			// }
+			// [DllImport ("libExpermimentalTests.dylib", EntryPoint = "_$s18ExpermimentalTests6OneIntV1XSivg")]
+			// public static extern nint XGetter (XXOpaque1 r);
+			//
+
+			var xxopName = new CSIdentifier ("XXOpaque1");
+			var selfID = (CSIdentifier)"self";
+			var thisSwiftDataPtrID = (CSIdentifier)"thisSwiftDataPtr";
+			var pID = (CSIdentifier)"p";
+			var getterID = (CSIdentifier)"Getter";
+			var xgetterID = (CSIdentifier)"XGetter";
+
+			var st = new CSStruct (CSVisibility.Public, xxopName);
+			st.Fields.Add (CSFieldDeclaration.FieldLine (CSSimpleType.NInt, "x0", value: null, CSVisibility.None));
+			var body = new CSCodeBlock ();
+			var fixedBlock = new CSFixedCodeBlock (CSSimpleType.ByteStar, thisSwiftDataPtrID,
+				new CSFunctionCall ("SwiftRuntimeLibrary.SwiftMarshal.StructMarshal.Marshaler.PrepareValueType", false, selfID), null);
+
+			fixedBlock.Add (CSVariableDeclaration.VarLine (pID, new CSCastExpression (new CSSimpleType ($"{xxopName.Name}*"), thisSwiftDataPtrID)));
+			fixedBlock.Add (CSReturn.ReturnLine (new CSFunctionCall (xgetterID.Name, false, new CSUnaryExpression (CSUnaryOperator.Indirection, pID))));
+			body.Add (fixedBlock);
+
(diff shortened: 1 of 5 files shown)


tests/tom-swifty-test/SwiftReflector/ExpermimentalTests.cs
```diff
@@ -0,0 +1,255 @@
...
+@frozen
+public struct OneInt {
+	public var X:Int
+	public init () { X = 17 }
+}
+";
+
+			// C# aux code:
+			// public struct XXOpaque1 {
+			// nint x0
```
mcumming, line 27: What is x0 in this case?

tests/tom-swifty-test/SwiftReflector/ExpermimentalTests.cs
```diff
@@ -0,0 +1,255 @@
...
+	public var X:Int
+	public init () { A = 7; X = 17 }
+}
+";
+
+			// C# aux code:
+			// public struct XXOpaque2 {
+			// nint x0
+			// nint x1
+			// public static unsafe nint Getter (OneInt self)
```
mcumming, line 87: What do x0 and x1 represent for the remainder of the test cases

stephen-hawley: In the current implementation of BTfS, the payload for value types is an array and we always pass it by reference. The reason is that the Swift ABI allows for value types to be passed by value that are up to 4 machine words (consuming 4 registers), but the standard ABI only allows for 2. This requires me to write adapters for all methods and functions that take value types. We don't like this because it's really inefficient. The runtime team is working on this so I can improve the marshaling. So, I made opaque structs that just take up space which I can use in pinvokes by blitting the payload array contents directly into them. I defined these two just for the sake of the test to answer the question "will this even work". Since my runtime currently stops at 2 machine words for pass by value, this is where I did the cutoff.
Make sense?

tests/tom-swifty-test/SwiftReflector/ExpermimentalTests.cs
```diff
@@ -0,0 +1,255 @@
...
+			// {
+			//    fixed (byte* thisSwiftDataPtr = SwiftRuntimeLibrary.SwiftMarshal.StructMarshal.Marshaler.PrepareValueType (self)) {
+			//        var p = (XXOpaque1 *)thisSwiftDataPtr;
+			//        return XGetter (*p);
+			// }
+			// [DllImport ("libExpermimentalTests.dylib", EntryPoint = "_$s18ExpermimentalTests6OneIntV1XSivg")]
+			// public static extern nint XGetter (XXOpaque1 r);
```
rolfbjarne, line 94:
```suggestion
			//        var p = (XXOpaque2 *)thisSwiftDataPtr;
			//        return XGetter (*p);
			// }
			// [DllImport ("libExpermimentalTests.dylib", EntryPoint = "_$s18ExpermimentalTests6OneIntV1XSivg")]
			// public static extern nint XGetter (XXOpaque2 r);
```

tests/tom-swifty-test/SwiftReflector/ExpermimentalTests.cs
```diff
@@ -0,0 +1,255 @@
...
```
rolfbjarne, line 155:
```suggestion
			//        var p = (XXOpaque2 *)thisSwiftDataPtr;
			//        return XGetter (*p);
			// }
			// [DllImport ("libExpermimentalTests.dylib", EntryPoint = "_$s18ExpermimentalTests6OneIntV1XSivg")]
			// public static extern nint XGetter (XXOpaque2 r);
```

tests/tom-swifty-test/SwiftReflector/ExpermimentalTests.cs
```diff
@@ -0,0 +1,255 @@
...
```
rolfbjarne, line 217:
```suggestion
			//        var p = (XXOpaque2 *)thisSwiftDataPtr;
			//        return XGetter (*p);
			// }
			// [DllImport ("libExpermimentalTests.dylib", EntryPoint = "_$s18ExpermimentalTests6OneIntV1XSivg")]
			// public static extern nint XGetter (XXOpaque2 r);
```

tests/tom-swifty-test/SwiftReflector/ExpermimentalTests.cs
```diff
@@ -0,0 +1,255 @@
...
+			// public struct XXOpaque1 {
+			// nint x0
+			// public static unsafe nint Getter (OneInt self)
+			// {
+			//    fixed (byte* thisSwiftDataPtr = SwiftRuntimeLibrary.SwiftMarshal.StructMarshal.Marshaler.PrepareValueType (self)) {
+			//        var p = (XXOpaque1 *)thisSwiftDataPtr;
+			//        return XGetter (*p);
+			// }
+			// [DllImport ("libExpermimentalTests.dylib", EntryPoint = "_$s18ExpermimentalTests6OneIntV1XSivg")]
+			// public static extern nint XGetter (XXOpaque1 r);
```
rolfbjarne, line 35: Declaring the argument as a `byte*` makes the code simpler:

```suggestion
			// public static unsafe nint Getter (OneInt self)
			// {
			//    fixed (byte* thisSwiftDataPtr = SwiftRuntimeLibrary.SwiftMarshal.StructMarshal.Marshaler.PrepareValueType (self)) {
			//        return XGetter (p);
			// }
			// [DllImport ("libExpermimentalTests.dylib", EntryPoint = "_$s18ExpermimentalTests6OneIntV1XSivg")]
			// public unsafe static extern nint XGetter (byte* r);
```

rolfbjarne (changes requested):

tests/tom-swifty-test/SwiftReflector/ExpermimentalTests.cs
stephen-hawley: But it's not a byte* this is pass by value not pass by reference, so I can't really do that.

rolfbjarne: Ah, I see, nvm.

rolfbjarne (approved):

PR #790, opened 2023-11-10
document the ABI conventions for swift-evolution

stephen-hawley:


docs/LibraryEvolution.md
```diff
@@ -0,0 +1,64 @@
...
+  is passed in the self register
+- instance values are passed by reference in the self register
+
+## value types not marked as frozen
+- return values are allocated by the caller and a pointer gets passed in to that space (note
+  that the size can and must be determined at runtime by the caller)
+- typical arguments are copied and passed by reference
+- inout arguments are passed by reference
+- instance values are passed in a copy passed in the self register
+- instance values are passed in the self register (no copy)
```
rolfbjarne, line 64: I'm assuming these two are passed by reference?

If so, just to make it explicit:

```suggestion
- instance values are passed by reference as a copy passed in the self register
- instance values are passed by reference in the self register (no copy)
```

rolfbjarne (approved):

Issue #793, opened 2023-11-17
[parser] InlineFunction is showing an weird parsing issue


stephen-hawley:
Running XmlReflectionTests.DynamicXmlTests.InlineFunction(Parser) is giving the errors:
```
line 8:11 missing {Binary_literal, Octal_literal, Decimal_literal, Pure_decimal_digits, Hexadecimal_literal} at 'b'
```

PR #795, opened 2023-11-20
Fixed a parsing error in Dyck expressions

stephen-hawley:
Added missing operators to the standard list to stop parse errors where there is inline code that uses these.
This fixes issue [793](https://github.com/xamarin/binding-tools-for-swift/issues/793).

As background, a Dyck expression is any expression which consists of arbitrary token that may be surrounded by balanced sets of a bracketing tokens. Inline functions and default values in Swift may have very complicated expressions, but we don't care at all about these, so we can get away with writing these of as a Dyck expression, of which the Swift language is a subset.

### DIFF
--- a/swiftinterfaceparser/SwiftInterface.g4
+++ b/swiftinterfaceparser/SwiftInterface.g4
@@ -297,7 +297,7 @@ dyckSubExpression :
 	| any_other_things_for_dyck_expression;
 
 any_other_things_for_dyck_expression :	
-	( OpDot | OpComma | OpColon | OpSemi | OpAssign | OpAt | OpPound | OpBackTick | OpQuestion | OpUnder)
+	( OpDot | OpComma | OpColon | OpSemi | OpAssign | OpAt | OpPound | OpBackTick | OpQuestion | OpUnder | OpPlus | OpMinus | OpAmp | OpBang | OpTilde | OpGreater | opGreater)
 	| arrow_operator
 	;
 


dustin-wojciechowski (approved):

rolfbjarne (approved):

swiftinterfaceparser/SwiftInterface.g4
```diff
@@ -297,7 +297,7 @@ dyckSubExpression :
 	| any_other_things_for_dyck_expression;
 
 any_other_things_for_dyck_expression :	
-	( OpDot | OpComma | OpColon | OpSemi | OpAssign | OpAt | OpPound | OpBackTick | OpQuestion | OpUnder)
+	( OpDot | OpComma | OpColon | OpSemi | OpAssign | OpAt | OpPound | OpBackTick | OpQuestion | OpUnder | OpPlus | OpMinus | OpAmp | OpBang | OpTilde | OpGreater | opGreater)
```
mcumming, line 300: Are there any other operators that should be in this list?

mcumming (approved):

swiftinterfaceparser/SwiftInterface.g4
stephen-hawley: Great question - I tried to name all the operators in the language with `Op` prefix. These are all of them. Swift has general operators as well which are it's own weird token syntax which should get picked up by the production rules under `dyckExcression` with the `literal` and `operator` rules.

PR #796, opened 2023-11-21
ideas on exposing C# to swift

stephen-hawley:


docs/FutureIdeas.md
```diff
@@ -66,3 +66,181 @@ and protocol witness tables, and leave all of this in a context object.
...
+attributes - we might have `[ExposeToSwift (swiftAttributes = "@available (iOS 10.0, macOS 10.12)\n@testable")]`
+It would probably also behoove us to provide a way to have a possibly different name for a type, method, or
+property in Swift than in C# in the event that we create a name collision that needs to be resolved.
+
+So we might have an implementation like this:
+```csharp
+public class ExposeToSwiftAttribute : Attribute {
+    public string SwiftAttributes { get; private set; }
+    public string SwiftName { get; private set; }
+    public ExposeToSwiftAttribte (string swiftAttributes = "", swiftName = "")
```
rolfbjarne, line 241:
```suggestion
    public ExposeToSwiftAttribute (string swiftAttributes = "", swiftName = "")
```

docs/FutureIdeas.md
```diff
@@ -66,3 +66,181 @@ and protocol witness tables, and leave all of this in a context object.
...
+There is one problem though - when you add new functionality to the the type, Swift will
+know nothing about it. If you have a dozen new C# subclasses of the original Swift class,
+you will not see the new functionality in Swift - they will all look like the singular
+Swift override.
+
+We can fix that - here's one way of doing that.
+Suppose we have an attribute named `[ExposeToSwift]` that can be applied to the type. That way if
+I write this in C#:
+```csharp
+[ExposeToSwift]
```
rolfbjarne, line 209: For Objective-C this happens automatically for all types that subclass NSObject - for Swift we could also do something similar, where we expose all types that subclass a Swift type.

docs/FutureIdeas.md
```diff
@@ -66,3 +66,181 @@ and protocol witness tables, and leave all of this in a context object.
...
+    public override init () { super.init () }
+}
+```
+And call this constructor instead of the `xam_sub_MontyWSMBool` constructor, then Swift will
+now see a class called `Bar` instead of a class called `xam_sub_MontyWSMBool`.
+
+This approach should also work for implementing protocols, including protocols with
+associated types to the degree that the C# type conformance matches any Swift requirements.
+
+In addition, we could allow `[ExposeToSwift]` to be applied to properties and methods as well.
```
rolfbjarne, line 225: We do this automatically for Objective-C: if a managed method or property in a subclass overrides an Objective-C method or property, it's exposed automatically.

IMHO this makes sense, because it seems more common to want to expose an overridden Swift method than not.

PR #799, opened 2023-11-29
add some basic parallelism to improve performance

stephen-hawley:
First cut on making aspects of BTfS parallelizable.

ErrorHandling and the *Inventory stack needed to be made thread-safe.

I made reflection and demangling async and changed how they're being invoked to run them in parallel.
Across the entire suit of unit tests, this improves performance by about 7%.

I'm fairly certain that I can make the C# code generation parallel as well - but I'll do that in a separate PR.

### DIFF
--- a/SwiftReflector/ErrorHandling.cs
+++ b/SwiftReflector/ErrorHandling.cs
@@ -11,6 +11,7 @@ using ObjCRuntime;
 
 namespace SwiftReflector {
 	public class ErrorHandling {
+		object messagesLock = new object ();
 		List<ReflectorError> messages;
 
 		public ErrorHandling ()
@@ -37,37 +38,61 @@ namespace SwiftReflector {
 
 		public void Add (ErrorHandling eh)
 		{
-			messages.AddRange (eh.messages);
+			lock (messagesLock) {
+				lock (eh.messagesLock) {
+					messages.AddRange (eh.messages);
+				}
+			}
 		}
 
 		public void Add (params ReflectorError [] errors)
 		{
-			messages.AddRange (errors);
+			lock (messagesLock) {
+				messages.AddRange (errors);
+			}
 		}
 
 		public void Add (Exception exception)
 		{
+			lock (messagesLock) {
 #if CRASH_ON_EXCEPTION
 			ExceptionDispatchInfo.Capture (exception).Throw ();
 #else
-			messages.Add (new ReflectorError (exception));
+				messages.Add (new ReflectorError (exception));
 #endif
+			}
 		}
 
 		public bool AnyMessages {
-			get { return messages.Count > 0; }
+			get {
+				lock (messagesLock) {
+					return messages.Count > 0;
+				}
+			}
 		}
 
 		public bool AnyErrors {
-			get { return messages.Any ((v) => !v.IsWarning);  }
+			get {
+				lock (messagesLock) {
+					return messages.Any ((v) => !v.IsWarning);
+				}
+			}
 		}
 
 		public int WarningCount {
-			get { return messages.Count ((v) => v.IsWarning); }
+			get {
+				lock (messagesLock) {
+					return messages.Count ((v) => v.IsWarning);
+				}
+			}
 		}
 
 		public int ErrorCount {
-			get { return messages.Count ((v) => !v.IsWarning); }
+			get {
+				lock (messagesLock) {
+					return messages.Count ((v) => !v.IsWarning);
+				}
+			}
 		}
 
 		public int Show (int verbosity)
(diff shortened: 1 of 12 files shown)


SwiftReflector/Inventory/ClassInventory.cs
```diff
@@ -22,12 +22,14 @@ public override void Add (TLDefinition tld, Stream srcStm)
 				return;
 			SwiftName className = ToClassName (tld);
 			SwiftClassName formalName = ToFormalClassName (tld);
-			ClassContents contents = null;
-			if (!values.TryGetValue (className, out contents)) {
-				contents = new ClassContents (formalName, sizeofMachinePointer);
-				values.Add (className, contents);
+			lock (valuesLock) {
```
mandel-macaque, line 25: Same as with the ErrorHandler, we could use https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-8.0 

Looking at the code on which you are locking, you could remove the need to the TryGetValue, use TryAdd which will add only if not present. If you were to update, TryUpdate would be the method you want. 


I don't know the code base, but you might be in a situation in which some Invetories are thread-safe and others are not.

SwiftReflector/Inventory/FunctionInventory.cs
```diff
@@ -22,12 +22,14 @@ public override void Add (TLDefinition tld, Stream srcStm)
...
-				overloads = new OverloadInventory (tlf.Name, sizeofMachinePointer);
-				values.Add (tlf.Name, overloads);
+			lock (valuesLock) {
+				OverloadInventory overloads = null;
+				if (!values.TryGetValue (tlf.Name, out overloads)) {
+					overloads = new OverloadInventory (tlf.Name, sizeofMachinePointer);
+					values.Add (tlf.Name, overloads);
+				}
+				overloads.Add (tlf, srcStm);
 			}
```
mandel-macaque, line 32: Same, which also means that changing the dictionary in the base type will make all of your subclasses thread-safe.

SwiftReflector/Inventory/Inventory.cs
```diff
@@ -8,6 +8,7 @@
 
 namespace SwiftReflector.Inventory {
 	public abstract class Inventory<T> {
+		protected object valuesLock = new object ();
 		protected Dictionary<SwiftName, T> values = new Dictionary<SwiftName, T> ();
```
mandel-macaque, line 12: A ConcurrentDict will make your life easier.

SwiftReflector/Inventory/ModuleInventory.cs
```diff
@@ -24,12 +25,14 @@ public class ModuleInventory : Inventory<ModuleContents> {
...
-			if (!values.TryGetValue (tld.Module, out module)) {
-				module = new ModuleContents (tld.Module, SizeofMachinePointer);
-				values.Add (tld.Module, module);
+			lock (valuesLock) {
+				ModuleContents module = null;
+				if (!values.TryGetValue (tld.Module, out module)) {
+					module = new ModuleContents (tld.Module, SizeofMachinePointer);
+					values.Add (tld.Module, module);
+				}
+				module.Add (tld, srcStm);
```
mandel-macaque, line 34: Same, all this methods are yy most of the time.

SwiftReflector/Inventory/ModuleInventory.cs
```diff
@@ -111,6 +114,23 @@ public static ModuleInventory FromFile (string pathToDynamicLibrary, ErrorHandli
...
+			if (streams.Count == 0)
+				return null;
+			var inventory = new ModuleInventory ();
+			try {
+				var tasks = streams.Select (stm => FromStreamIntoAsync (stm, inventory, errors, stm.Name));
+				await Task.WhenAll (tasks);
+			} finally {
+				foreach (var stm in streams) {
+					stm.Dispose ();
+				}
```
mandel-macaque, line 129: The Reactive Extensions (https://github.com/dotnet/reactive)  provides a nice way to deal with a collection of IDisposables: https://learn.microsoft.com/en-us/previous-versions/dotnet/reactive-extensions/hh229918(v=vs.103)

I don't know how you feel about adding it as a dependency "System.Reactive" but will probably help.

mandel-macaque (commented):
I won't block the code, but there are a low of areas in which we could use the BCL concurren classes to make things work better. 

We are also using Threads for IO bound work, while indeed will make the code async, that is nor the proper way to work since we are using CPU threads for stuff that the IO can deal with better.

SwiftReflector/Inventory/ClassInventory.cs
stephen-hawley: opened an issue for it. https://github.com/xamarin/binding-tools-for-swift/issues/800

SwiftReflector/ErrorHandling.cs
```diff
@@ -11,6 +11,7 @@
 
 namespace SwiftReflector {
 	public class ErrorHandling {
+		object messagesLock = new object ();
```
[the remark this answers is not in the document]
stephen-hawley: opened an issue for it https://github.com/xamarin/binding-tools-for-swift/issues/801

PR #818, opened 2024-02-08
remove nullable warnings

stephen-hawley:
Added postfix unary operators to dynamo.
Changed all cases of calling `GetNSObject` and `CSObjectForSwiftObject` to have `!` after the call.

Why not inject a null check and exception? Because if these calls return `null` you've got bigger problems on your hands and the runtime will throw and exception for you.

### DIFF
--- a/Dynamo/Dynamo.CSLang/CSUnaryExpression.cs
+++ b/Dynamo/Dynamo.CSLang/CSUnaryExpression.cs
@@ -12,8 +12,13 @@ namespace Dynamo.CSLang {
 		}
 		protected override void LLWrite (ICodeWriter writer, object o)
 		{
-			writer.Write (string.Format (Operation == CSUnaryOperator.Ref || Operation == CSUnaryOperator.Out ? "{0} " : "{0}", OperatorToString (Operation)), true);
-			Expr.WriteAll (writer);
+			if (IsPostfix (Operation)) {
+				Expr.WriteAll (writer);
+				writer.Write (string.Format (Operation == CSUnaryOperator.Ref || Operation == CSUnaryOperator.Out ? "{0} " : "{0}", OperatorToString (Operation)), true);
+			} else {
+				writer.Write (string.Format (Operation == CSUnaryOperator.Ref || Operation == CSUnaryOperator.Out ? "{0} " : "{0}", OperatorToString (Operation)), true);
+				Expr.WriteAll (writer);
+			}
 		}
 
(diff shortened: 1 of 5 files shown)


Dynamo/Dynamo.CSLang/CSUnaryExpression.cs
```diff
@@ -12,8 +12,13 @@ public CSUnaryExpression (CSUnaryOperator op, ICSExpression expr)
...
 		{
-			writer.Write (string.Format (Operation == CSUnaryOperator.Ref || Operation == CSUnaryOperator.Out ? "{0} " : "{0}", OperatorToString (Operation)), true);
-			Expr.WriteAll (writer);
+			if (IsPostfix (Operation)) {
+				Expr.WriteAll (writer);
+				writer.Write (string.Format (Operation == CSUnaryOperator.Ref || Operation == CSUnaryOperator.Out ? "{0} " : "{0}", OperatorToString (Operation)), true);
+			} else {
+				writer.Write (string.Format (Operation == CSUnaryOperator.Ref || Operation == CSUnaryOperator.Out ? "{0} " : "{0}", OperatorToString (Operation)), true);
+				Expr.WriteAll (writer);
+			}
```
mandel-macaque, line 21: When is OperatorToString (Operation) used? Can it return the Operator with the space directly rather than copy pasting? 

```csharp
static string OperatorToString (CSUnaryOperator op)
		{
			switch (op) {
			case CSUnaryOperator.At:
				return "@";
			case CSUnaryOperator.BitNot:
				return "~";
			case CSUnaryOperator.Neg:
				return "-";
			case CSUnaryOperator.Not:
				return "!";
			case CSUnaryOperator.Out:
				return "out ";
			case CSUnaryOperator.Pos:
				return "+";
			case CSUnaryOperator.Ref:
				return "ref ";
			case CSUnaryOperator.AddressOf:
				return "&";
			case CSUnaryOperator.Indirection:
				return "*";
			case CSUnaryOperator.Await:
				return "await ";
			case CSUnaryOperator.PostBang:
				return "!";
			case CSUnaryOperator.Question:
				return "?";
			default:
				throw new ArgumentOutOfRangeException (nameof(op));
			}
		}
```

mandel-macaque (changes requested):

rolfbjarne (approved):

mandel-macaque (approved):

PR #822, opened 2024-02-12
apparently, this wants a "/private" in the front.

stephen-hawley:
FYI - Path.Combine doesn't correctly add "/private" on the front.

### DIFF
--- a/tests/tom-swifty-test/PosixUtilsTests.cs
+++ b/tests/tom-swifty-test/PosixUtilsTests.cs
@@ -40,7 +40,8 @@ namespace tomwiftytest {
 		public void RelativePathsInSymlinks1 ()
 		{
 			using (var dir = new DisposableTempDirectory ()) {
-				var expectedPath = "/private" + Path.GetDirectoryName (Path.GetDirectoryName (Path.GetDirectoryName (dir.DirectoryPath)));
+				var sourcePath = PathWithoutLeadingSlash (Path.GetDirectoryName (Path.GetDirectoryName (Path.GetDirectoryName (dir.DirectoryPath))));
+				var expectedPath = Path.Combine ("/private", sourcePath);
 
 				var link1 = Path.Combine (dir.DirectoryPath, "link1");
 				symlink ("../../..", link1);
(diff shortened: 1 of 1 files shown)


dustin-wojciechowski (approved):

tests/tom-swifty-test/PosixUtilsTests.cs
```diff
@@ -40,11 +40,12 @@ public void DotDotWorks1 ()
 		public void RelativePathsInSymlinks1 ()
 		{
 			using (var dir = new DisposableTempDirectory ()) {
-				var expectedPath = Path.GetDirectoryName (Path.GetDirectoryName (Path.GetDirectoryName (dir.DirectoryPath)));
+				var expectedPath = "/private" + Path.GetDirectoryName (Path.GetDirectoryName (Path.GetDirectoryName (dir.DirectoryPath)));
```
mandel-macaque, line 43: That is strange, look at the following example: 
Screenshot 2024-02-12 at 16 21 25

This is running the mono RELP but I'd be surprised that dotnet does to do the right things. Can you show us what was the code you tested?

mandel-macaque (changes requested):

tests/tom-swifty-test/PosixUtilsTests.cs
stephen-hawley: I tried:
```
Path.Combine ("/private", Path.GetDirectoryName (Path.GetDirectoryName (Path.GetDirectoryName (dir.DirectoryPath))));
```
and
```
Path.Combine ("/private/", Path.GetDirectoryName (Path.GetDirectoryName (Path.GetDirectoryName (dir.DirectoryPath))));
```

haritha-mohan (approved):

tests/tom-swifty-test/PosixUtilsTests.cs
rolfbjarne: If the second path is an absolute path, then the second path is returned as-is.

```cs
Path.Combine ("/a", "/b"); // prints "/b"
Path.Combine ("/a", "b"); // prints "/a/b"
```

Also note that the `/private` prefix is because `Path.GetTempPath ()` returns something like `/var/folders/..`, where `/var ` is a symlink to `/private/var`:

```shel
$ ls -lad /var
lrwxr-xr-x@ 1 root  wheel  11 Jan 12 08:19 /var -> private/var
```

so maybe a better fix would be to do something like this?

```cs
				var realDirectoryPath = PosixHelpers.RealPath (dir.DirectoryPath);
				var expectedPath = Path.GetDirectoryName (Path.GetDirectoryName (Path.GetDirectoryName (realDirectoryPath)));
```

rolfbjarne (approved):

rolfbjarne (approved):

dalexsoto (approved):

mandel-macaque (approved):

PR #830, opened 2024-02-15
fix rt unmanagedcallers

stephen-hawley:
Tests were failing on this because the compiler doesn't catch this error.
Also changed this to int instead of byte to avoid the casting.

### DIFF
--- a/SwiftRuntimeLibrary/SwiftEquatableProxy.cs
+++ b/SwiftRuntimeLibrary/SwiftEquatableProxy.cs
@@ -22,7 +22,7 @@ namespace SwiftRuntimeLibrary {
 
 		struct Equatable_xam_vtable {
 			public delegate bool Delfunc0 (IntPtr one, IntPtr two);
-			public unsafe delegate *unmanaged<IntPtr, IntPtr, byte> func0;
+			public unsafe delegate *unmanaged<IntPtr, IntPtr, int> func0;
 		}
 
 		static Equatable_xam_vtable vtableIEquatable;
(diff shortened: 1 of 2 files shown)


mcumming (approved):

haritha-mohan (approved):

dalexsoto (commented):
don't you want `byte` instead of an `int`?

SwiftRuntimeLibrary/SwiftEquatableProxy.cs
```diff
@@ -32,15 +32,15 @@ static SwiftEquatableProxy ()
 		}
 
 		[UnmanagedCallersOnly]
-		static byte EqFunc (IntPtr oneptr, IntPtr twoptr)
+		static int EqFunc (IntPtr oneptr, IntPtr twoptr)
```
rolfbjarne, line 35: As @dalexsoto says, you need to match the size of the native type, otherwise you can end up reading random bits and you'll see "true" when the native code meant "false" (and yes, this happens, I've run into it in the past).

rolfbjarne (changes requested):

SwiftRuntimeLibrary/SwiftEquatableProxy.cs
stephen-hawley: The code that uses this always masks off the 0. How would that be an issue?
I mean, I could make everything return `nint` or `nuint` but I figured this is likely to aot into nicer code.

rolfbjarne: Ah, I see, returning an `int` from a function when the caller expects a `bool` should be fine (the opposite - expecting an `int` in the caller when the callee returns a `byte` or `bool` - is not, but that's not the case here).

rolfbjarne (approved):

PR #833, opened 2024-02-29
Refactor vtable handling

stephen-hawley:
This is...not small.

The changes consist of 3 major elements:
1 - changes to Dynamo for writing classes/methods that are `unsafe`
2 - refactoring NewClassCompiler such that nearly all auxiliary classes live in `VTableDetails` and handling of auxiliary classes is uniform.
3 - remove all vtable generation code to `VTableDetails` which now writes vtables such that there is a "local" vtable which contains managed delegates that call local receivers and there is a swift vtable which contains unmanaged delegates which call receivers that retrieve the local vtable and trampoline out to them.

The last element is the most important. What this does is allows the unmanaged delegates to live in their own class and compile correctly with `[UnmanagedCallersOnly]` set.

A lot of this is non-obvious code, so I tried to document VTableDetails to make it clear what's going on.

A couple notes:
There will be a local/swift vtable for every type that needs a vtable. The local vtable is technically not needed for non-generic classes, but that could be an optimization for another day.
TIL that you can't have pointer types in generics eg, `Func<ExistentialContainer *>` this required me using `delegate` types instead which are quite happy to have pointers, but at the same time `delegate *unmanaged<>` doesn't like pointer types for the same reason so I had to inject some ham-handed casting to turn pointer types into `IntPtr` and cast them back.

### DIFF
--- a/SwiftReflector/VTableDetails.cs
+++ b/SwiftReflector/VTableDetails.cs
@@ -309,8 +309,7 @@ namespace SwiftReflector
 
 			var pl = delType.Parameters;
 			var plWithoutGCPtr = new CSParameterList (pl.Skip (1));
-			var usedIDs = new List<string> ();
-			usedIDs.AddRange (pl.Select (p => p.Name.Name));
+			var usedIDs = new List<string> (pl.Select (p => p.Name.Name));
 
 			var marshal = new MarshalEngineCSafeSwiftToCSharp (use, usedIDs, typeMapper);
 
@@ -392,9 +391,8 @@ namespace SwiftReflector
 			var returnType = funcDecl.IsGetter ? delType.Type : CSSimpleType.Void;
 
 			var pl = delType.Parameters;
-			var plWithoutGCPtr = new CSParameterList (pl.GetRange (1, pl.Count - 1));
-			var usedIDs = new List<string> ();
-			usedIDs.AddRange (pl.Select (p => p.Name.Name));
+			var plWithoutGCPtr = new CSParameterList (pl.Skip (1));
+			var usedIDs = new List<string> (pl.Select (p => p.Name.Name));
 
 			var body = new CSCodeBlock ();
 
@@ -431,8 +429,7 @@ namespace SwiftReflector
 
 			var pl = delType.Parameters;
 			var plWithoutGCPtr = new CSParameterList (pl.Skip (1));
-			var usedIDs = new List<string> ();
-			usedIDs.AddRange (pl.Select (p => p.Name.Name));
+			var usedIDs = new List<string> (pl.Select (p => p.Name.Name));
 
 			var body = new CSCodeBlock ();
 
--- a/SwiftReflector/VTableDetails.cs
+++ b/SwiftReflector/VTableDetails.cs
@@ -1,0 +1,490 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Runtime.Intrinsics.X86;
+using Dynamo;
+using Dynamo.CSLang;
+using SwiftReflector.SwiftXmlReflection;
+using SwiftReflector.TypeMapping;
+using SwiftRuntimeLibrary;
+using SwiftRuntimeLibrary.SwiftMarshal;
+
+namespace SwiftReflector
+{
+	public class VTableDetails
+	{
+		int currentVTEntry = 0;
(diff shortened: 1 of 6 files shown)


rolfbjarne (approved):

SwiftReflector/VTableDetails.cs
```diff
@@ -0,0 +1,490 @@
...
+		public CSMethod ImplementVirtualSubscriptStaticReceiver (CSType thisType, string csProxyName, CSDelegateTypeDecl delType, CSUsingPackages use,
+						FunctionDeclaration funcDecl, CSProperty prop, CSMethod protoListMethod, TypeMapper typeMapper, bool isObjC,
+						Func<int, int, string> genericRenamer = null, bool hasAssociatedTypes = false)
+		{
+			var returnType = funcDecl.IsSubscriptGetter ? delType.Type : CSSimpleType.Void;
+
+			var pl = delType.Parameters;
+			var plWithoutGCPtr = new CSParameterList (pl.Skip (1));
+			var usedIDs = new List<string> ();
+			usedIDs.AddRange (pl.Select (p => p.Name.Name));
```
mandel-macaque, line 435: The List<T> class already has a constructor that takes a constructor that does the init and adds the range: https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.-ctor?view=net-8.0#system-collections-generic-list-1-ctor(system-collections-generic-ienumerable((-0)))

SwiftReflector/VTableDetails.cs
```diff
@@ -0,0 +1,490 @@
...
+		}
+
+		public CSMethod ImplementVirtualPropertyStaticReceiver (CSType thisType, string csProxyName, CSDelegateTypeDecl delType,
+					       CSUsingPackages use, FunctionDeclaration funcDecl, CSProperty prop, CSMethod protoListMethod,
+					       TypeMapper typeMapper, bool isObjC, bool hasAssociatedTypes, Func<int, int, string> genericRenamer = null)
+		{
+			var returnType = funcDecl.IsGetter ? delType.Type : CSSimpleType.Void;
+
+			var pl = delType.Parameters;
+			var plWithoutGCPtr = new CSParameterList (pl.GetRange (1, pl.Count - 1));
```
mandel-macaque, line 395: Careful, I don't know if possible but if p1.Count is 0, GetRange will throw an exception: https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.getrange?view=net-8.0#exceptions

SwiftReflector/VTableDetails.cs
```diff
@@ -0,0 +1,490 @@
...
+			var usedIDs = new List<string> ();
+			usedIDs.AddRange (pl.Select (p => p.Name.Name));
```
mandel-macaque, line 397: Can be merged in the constructor (sorry I do the reviews sometimes from bottom to top, my brain is weird).

SwiftReflector/VTableDetails.cs
```diff
@@ -0,0 +1,490 @@
...
+			// {
+			//      var gch = GCHandle.FromIntPtr (gcPtr);
+			//      var localVT = (LocalVTable)gch.Target;
+			//      [return] localVT.Func0(args);
+			// }
+
+			var pl = delType.Parameters;
+			var plWithoutGCPtr = new CSParameterList (pl.Skip (1));
+			var usedIDs = new List<string> ();
+			usedIDs.AddRange (pl.Select (p => p.Name.Name));
```
mandel-macaque, line 313: Can be unified.

SwiftReflector/VTableDetails.cs
```diff
@@ -0,0 +1,490 @@
...
+		{
+			yield return PInvokes;
+			if (HasActualVTable) {
+				yield return LocalCSVTable;
+				yield return UnmanagedReceivers;
+				yield return SwiftVtable;
+			}
+		}
+
+		public bool HasActualVTable => currentVTEntry > 0;
```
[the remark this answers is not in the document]
stephen-hawley: No. It will only get called after the class was processed.

SwiftReflector/VTableDetails.cs
stephen-hawley: All the parameter lists that make to this point will have the GC Handle as the first argument. Although I think I'm going to change that to `Skip (1)` since it reads better.

PR #837, opened 2024-03-14
first swath of naming utility refactoring.

stephen-hawley:
Naming of various things in BTfS is spread all over the place. I'm trying to centralize it so it's easier to refactor.

This is step 1 of n - and in this case I've created a new class for creating unique names which also includes the ability to abide by scoping. This code is not used...yet. The general approach for this is to use a stack of hash sets. When you enter a new scope the top-most hashset is used to populate the new scope. Here's the thinking on that: The contents of the hash sets will be relatively small: on the order of 10 items (ie, no fewer than 1 and no more than 100, but typically 6-8). Scopes will typically be exited and entered on function boundaries and I don't anticipate having to create new ones for control structures as these tend to nest very linearly. Checking for existing symbols should be cheaper than it was before and since that is the most common operation, I felt this was better than walking the stack of scopes, but PRs are welcome if you disagree.

The general philosophy here is to use static methods when things are stateless and classes when things require state.

I've also moved `SanitizeIdentifier` and related code from `TypeMapper` and `CleanseOperatorName` from `MethodWrapping` and put them into `CSSafeNaming`.

### DIFF
--- a/SwiftReflector/Naming/ScopedNaming.cs
+++ b/SwiftReflector/Naming/ScopedNaming.cs
@@ -11,9 +11,9 @@ namespace SwiftReflector.Naming
 		string autoSymPrefix;
 		Stack<HashSet<string>> names = new Stack<HashSet<string>> ();
 
-		public ScopedNaming (string? autoSymPrefix = null)
+		public ScopedNaming (string autoSymPrefix = "_symbol")
 		{
-			this.autoSymPrefix = string.IsNullOrEmpty (autoSymPrefix) ? "_symbol" : autoSymPrefix;
+			this.autoSymPrefix = autoSymPrefix;
 			names.Push (new HashSet<string> ());
 		}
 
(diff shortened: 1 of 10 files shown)


SwiftReflector/Naming/ScopedNaming.cs
```diff
@@ -0,0 +1,71 @@
...
+
+		public void EnterScope ()
+		{
+			names.Push (new HashSet<string> (names.Peek ()));
+		}
+
+		public void ExitScope ()
+		{
+			if (names.Count == 0)
+				throw new Exception ("exited top level scope - that's bad");
```
mcumming, line 28: Is this visible to the customer? If it is, it should be more informative

mcumming (approved):

SwiftReflector/Naming/ScopedNaming.cs
stephen-hawley: By being public, it is, in theory, visible to the customer but in practice it will only be used by BTfS (or its descendants).
So yeah, everything could be internal, but it makes testing harder.

rolfbjarne (approved):

